PR #7, opened 2022-03-03
feat: click on header name opens dimension modal

edoardo:
### Key features

1. click on column header opens the dimension modal

---

### Description

Similarly to the layout chip, clicking the column header in the data table opens the modal for editing the dimension.

---

### Screencast

https://user-images.githubusercontent.com/150978/156581522-5b5772f6-aa2d-4368-83f4-2680ed80d4da.mov


src/components/Visualization/Visualization.js
```diff
@@ -166,18 +174,19 @@ export const Visualization = ({
                 postfix = `${i18n.t('most recent')} ${header.stageOffset}`
```
martinkrulltott, line 174: why isnt this string using i18n interpolation?

src/components/Visualization/Visualization.js
```diff
@@ -153,6 +154,13 @@ export const Visualization = ({
     }
 
     const formatCellHeader = (header) => {
+        let headerName = header.column
+        let dimensionId = header.name
+
+        const match = Object.entries(headersMap).find(
```
martinkrulltott, line 160: Im a bit confused about this use of `Object.entries`, since it looks like only `match[0]` (key) is used below. 
If you specifically only need the key in the end, then you might as well do `Object.keys(headersMap).find(key => headersMap[key].name === header.name)`.

janhenrikoverland (approved):

PR #11, opened 2022-03-04
fix: chip label with stage name and correct suffix

jenniferarnesen:
Implements [TECH-1009](https://jira.dhis2.org/browse/TECH-1009)

---

### Key features

1. Add stageName to the chip label in cases where multiple data elements have the same name. This is the same functionality used by ProgramDimensionsList (code is duplicated for now, but will consider refactoring when adding stageName to modal and the visualization columns)
2. Style the chip such that the suffix showing #selected/conditions always shows in full, and the name-stageName part being truncated with overflow ellipsis.
3. move TooltipContent->renderConditions to function getConditions in modules/conditions. This was so that the functionality could be shared with the chip label component (ChipBase)
4. refactor: switch DefaultAxis to use useSelector/useDispatch instead of connect. This resulted in `renderChipsSelector` being moved to reducers/ui.js

---

### Screenshots

Remember to look at the text on the chip. The tooltip is shown so that you can compare what is on the chip to the actual items/conditions that were chosen

Display stage name on the chip in cases where there are duplicate data element names in the layout:
 
image

Number-legend set with 3 selected options
image

Regular number field with 2 conditions:
image

Number -option set with 4 selected options
image

Multiple types with nothing selected (take note of "Organisation unit" and "Event date":
image

Yes-only with 1 option selected
image

Yes-only with both options selected
image

Yes-only with no options selected
image

Boolean with 2 options selected
image

Boolean with all options selected
image

Boolean with no options selected
image

### DIFF
--- a/src/components/Layout/ChipBase.js
+++ b/src/components/Layout/ChipBase.js
@@ -12,6 +12,9 @@ import { DIMENSION_TYPE_PERIOD } from '../../modules/dimensionConstants.js'
 import { DimensionIcon } from '../MainSidebar/DimensionItem/DimensionIcon.js'
 import styles from './styles/Chip.module.css'
 
+const VALUE_TYPE_TRUE_ONLY_NUM_OPTIONS = 2
+const VALUE_TYPE_BOOLEAN_NUM_OPTIONS = 3
+
 // Presentational component used by dnd - do not add redux or dnd functionality
 
 const getChipSuffix = ({
@@ -39,8 +42,10 @@ const getChipSuffix = ({
     const conds = getConditions({ conditions, metadata, dimension })
 
     if (
-        (valueType === VALUE_TYPE_TRUE_ONLY && conds.length === 2) ||
-        (valueType === VALUE_TYPE_BOOLEAN && conds.length === 3)
+        (valueType === VALUE_TYPE_TRUE_ONLY &&
+            conds.length === VALUE_TYPE_TRUE_ONLY_NUM_OPTIONS) ||
+        (valueType === VALUE_TYPE_BOOLEAN &&
+            conds.length === VALUE_TYPE_BOOLEAN_NUM_OPTIONS)
     ) {
         return `: ${all}`
     }
--- a/src/components/Layout/ChipBase.js
+++ b/src/components/Layout/ChipBase.js
@@ -67,14 +67,19 @@ const getChipSuffix = ({
 }
 
 export const ChipBase = (props) => {
-    const { name, dimensionType } = props.dimension
+    const { name, dimensionType, stageName } = props.dimension
     return (
         <div className={cx(styles.chipBase)}>
             <div className={styles.leftIcon}>
                 <DimensionIcon dimensionType={dimensionType} />
             </div>
-            <span className={styles.label}>{name}</span>
-            <span className={styles.label}>{getChipSuffix(props)}</span>
+            <span className={styles.label}>
+                <span className={styles.primary}>{name}</span>
+                {stageName && (
+                    <span className={styles.secondary}>{stageName}</span>
+                )}
+            </span>
+            <span className={styles.suffix}>{getChipSuffix(props)}</span>
         </div>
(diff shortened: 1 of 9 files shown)


src/components/Layout/ChipBase.js
```diff
@@ -1,48 +1,94 @@
...
+            dimensionType === DIMENSION_TYPE_PERIOD) &&
+        !items.length
+    ) {
+        return ''
+    }
+
+    const all = i18n.t('all')
+
+    if (!conditions.condition && !conditions.legendSet && !items.length) {
+        return `: ${all}`
```
martinkrulltott, line 39: For consistency, might be best to either stick to the `suffix =` path or the direct return path (e.g. `return : '${all}'` everywhere).
So you could either move the `let suffix...` up higher and use that here.. or remove the `let suffix...` part below and just do a direct return instead.

src/components/Layout/ChipBase.js
```diff
@@ -1,48 +1,94 @@
...
+            <div className={styles.leftIcon}>
+                <DimensionIcon dimensionType={dimensionType} />
+            </div>
+            <span className={styles.label}>
+                <span className={styles.primary}>{name}</span>
+                {stageName && (
+                    <span className={styles.secondary}>{stageName}</span>
+                )}
+            </span>
+            <span className={styles.suffix}>{getChipSuffix(props)}</span>
```
martinkrulltott, line 87: Unless there's a specific reason for keeping the `getChipSuffix` function outside of the `ChipBase` scope, we might as well just include inside `ChipBase` and get the props directly from `ChipBase` instead of passing them in (then you can do the regular props destruction in the input params as we usually do).

`export const ChipBase = ({foo, bar, baz}) => { const getChipSuffix = () => ...`

src/components/Layout/ChipBase.js
```diff
@@ -1,48 +1,94 @@
...
-)
+    )
+}
 
 ChipBase.propTypes = {
-    dimensionName: PropTypes.string,
-    dimensionType: PropTypes.string,
-    items: PropTypes.array,
-    numberOfConditions: PropTypes.number,
+    dimension: PropTypes.object,
```
martinkrulltott, line 93: missing prop types for `items, conditions, metadata`

jenniferarnesen: oops how did that sneak by

src/components/Layout/DefaultLayout/DefaultAxis.js
```diff
@@ -2,50 +2,115 @@ import { useDroppable } from '@dnd-kit/core'
...
+    const programStageNames = useMemo(
+        () =>
+            program?.programStages?.reduce((acc, stage) => {
+                acc.set(stage.id, stage.name)
+                return acc
+            }, new Map()),
+        [program]
+    )
+
+    const dimensions = getDimensionsWithStageName({
```
martinkrulltott, line 106: same as above, any particular reason for keeping this fn outside of the `DefaultAxis` scope or can it just be put inside the scope so e.g. `metadata` can be used directly instead of first being defined above and then be passed in?

src/reducers/ui.js
```diff
@@ -360,12 +361,28 @@ export const sGetDimensionIdsFromLayout = (state) =>
         []
     )
 
+export const sGetUiDimensionIdsForLayoutAxis = (state, axisId) =>
```
martinkrulltott, line 364: rename to fit in to the other selectors?
```suggestion
export const sGetUiDimensionIdsByAxisId = (state, axisId) =>
```

src/components/Layout/ChipBase.js
jenniferarnesen: went with "return"

martinkrulltott (approved):

src/components/Layout/ChipBase.js
jenniferarnesen: ok

PR #43, opened 2022-03-16
fix: sort out rule-of-hook lint error

janhenrikoverland:


jenniferarnesen:
PR 44 implemented instead

jenniferarnesen:
PR 44 implemented instead

PR #56, opened 2022-03-22
fix: make URL values in datatable clickable TECH-1054

edoardo:
Implements [TECH-1054](https://dhis2.atlassian.net/browse/TECH-1054)

---

### Key features

1. make URLs clickable in datatable

---

### Description

Row values of `URL` type should be rendered as `<a>` tags to make them clickable and open in a new tab.

---

### Screenshots

Screenshot 2022-03-22 at 10 14 08

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -130,6 +130,12 @@ export const Visualization = ({
                     ? 'yyyy-MM-DD hh:mm'
                     : 'yyyy-MM-DD'
             )
+        } else if (header?.valueType === 'URL') {
+            return (
+                <a href={value} target="_blank" rel="noreferrer">
+                    {value}
+                </a>
+            )
         } else {
             return formatValue(value, header?.valueType || 'TEXT', {
                 digitGroupSeparator: visualization.digitGroupSeparator,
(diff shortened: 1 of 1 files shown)


src/components/Visualization/Visualization.js
```diff
@@ -130,6 +130,12 @@ export const Visualization = ({
                     ? 'yyyy-MM-DD hh:mm'
                     : 'yyyy-MM-DD'
             )
+        } else if (header?.valueType === 'URL') {
```
martinkrulltott, line 133:
```suggestion
        } else if (header?.valueType === VALUE_TYPE_URL) {
```

src/components/Visualization/Visualization.js
```diff
@@ -130,6 +130,12 @@ export const Visualization = ({
...
+            return (
+                <a href={value} target="_blank" rel="noreferrer">
+                    {value}
+                </a>
+            )
         } else {
             return formatValue(value, header?.valueType || 'TEXT', {
```
martinkrulltott, line 140:
```suggestion
            return formatValue(value, header?.valueType || VALUE_TYPE_TEXT, {
```

src/components/Visualization/Visualization.js
```diff
@@ -130,6 +130,12 @@ export const Visualization = ({
                     ? 'yyyy-MM-DD hh:mm'
```
martinkrulltott, line 130: Change `'DATE'` to `VALUE_TYPE_DATE` above

martinkrulltott (approved):
LGTM, but it can be improved by using the consts instead of strings for the value types

PR #58, opened 2022-03-23
fix: interpretations fixes TECH-1055

edoardo:
Implements [TECH-1055](https://dhis2.atlassian.net/browse/TECH-1055)

---

### Key features

1. add sharing icon in Interpretation toolbar
2. sort interpretations within a date in descending order
3. “Write a reply” instead of “Add a comment”
4. “Post reply” instead of “Save comment”
5. fix mentioning api request
6. spacing between interpretations thread and right margin

---

### Screenshots

1. Interpretations sorted descending:
Screenshot 2022-03-23 at 12 18 07

2. Sharing icon:
Screenshot 2022-03-23 at 12 16 36

3. 4. Write a reply/Post reply:
Screenshot 2022-03-23 at 12 19 12

5. User lookup request:
Screenshot 2022-03-23 at 12 20 37
Screenshot 2022-03-23 at 12 20 50

6. Right spacing:
Screenshot 2022-03-23 at 15 07 11

### DIFF
--- a/src/components/Interpretations/InterpretationsUnit/InterpretationList.js
+++ b/src/components/Interpretations/InterpretationsUnit/InterpretationList.js
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { Interpretation } from '../common/index.js'
 
-const sortByDateProp = (a, b) => {
+const sortByCreatedDateDesc = (a, b) => {
     const dateA = a.created
     const dateB = b.created
 
@@ -55,7 +55,7 @@ export const InterpretationList = ({
                         </div>
                         <ol className="interpretation-list">
                             {interpretationsByDate[date]
-                                .sort(sortByDateProp)
+                                .sort(sortByCreatedDateDesc)
                                 .map((interpretation) => (
                                     <Interpretation
                                         key={interpretation.id}
--- a/src/components/Interpretations/InterpretationsUnit/InterpretationList.js
+++ b/src/components/Interpretations/InterpretationsUnit/InterpretationList.js
@@ -9,10 +9,10 @@ const sortByDateProp = (a, b) => {
     const dateB = b.created
 
     if (dateA < dateB) {
-        return -1
+        return 1
     }
     if (dateA > dateB) {
-        return 1
+        return -1
     }
     return 0
 }
--- a/src/components/Interpretations/common/Interpretation/Interpretation.js
+++ b/src/components/Interpretations/common/Interpretation/Interpretation.js
@@ -1,5 +1,12 @@
 import i18n from '@dhis2/d2-i18n'
-import { Button, IconReply16, IconThumbUp16, IconEdit16 } from '@dhis2/ui'
+import {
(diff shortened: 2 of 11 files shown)


src/components/Interpretations/InterpretationsUnit/InterpretationList.js
```diff
@@ -9,10 +9,10 @@ const sortByDateProp = (a, b) => {
     const dateB = b.created
 
     if (dateA < dateB) {
```
jenniferarnesen, line 11: Maybe this function could be named `sortByCreatedDateDesc` ?

src/components/Interpretations/common/Interpretation/Interpretation.js
```diff
@@ -38,7 +47,8 @@ export const Interpretation = ({
             created={interpretation.created}
             id={interpretation.id}
             username={interpretation.user.displayName}
-            onClick={disabled ? undefined : onClick}
+            // XXX showSharingDialog is needed here to avoid the interpretations view to open when the SharingDialog is closed
```
jenniferarnesen, line 50: I think it's safe to remove this comment. The code is pretty self explanatory.

src/components/Interpretations/common/Interpretation/InterpretationUpdateForm.js
```diff
@@ -48,7 +49,9 @@ export const InterpretationUpdateForm = ({
                     disabled={loading}
                     errorText={errorText}
                 />
-                <InterpretationSharingLink id={id} type="interpretation" />
+                {showSharingLink && (
+                    <InterpretationSharingLink id={id} type="interpretation" />
```
jenniferarnesen, line 53: I'm kind of confused about the `type` property. This gets passed to SharingDialog, right? I don't find that anywhere in the SharingDialog code, but maybe I'm looking in the wrong place? The types are found are: 
https://github.com/dhis2/ui/blob/master/components/sharing-dialog/src/sharing-dialog.js#L194

edoardo: Yes, I need to merge this: https://github.com/dhis2/ui/pull/1007

src/components/Interpretations/InterpretationsUnit/InterpretationList.js
edoardo: Renamed.

src/components/Interpretations/common/Interpretation/Interpretation.js
edoardo: I don't see how the comment is wrong, which makes me think it is needed then 😅

src/components/Interpretations/common/Interpretation/InterpretationUpdateForm.js
jenniferarnesen: Ah, I see. Is there any way to export those type constants so they can be used here rather than using a string? `type={INTERPRETATION}`

PR #61, opened 2022-03-24
fix: reduce unnecessary rerenders by not creating new objects and arrays in selectors

jenniferarnesen:
Implements [TECH-990](https://dhis2.atlassian.net/browse/TECH-990)

### Key features

1. Define constants for empty objects and arrays so that we're not creating a new one with each call to the selector
2. Use PointerSensor instead of MouseSensor so dnd will work on more devices (also, PointerSensor is the default for dnd-kit)
3. useMemo and useCallback and React.memo in certain cases - led to a big reduction in renders of those components, although no real performance gain was seen.


jenniferarnesen:
Not doing this - no gain in performance

PR #62, opened 2022-03-25
fix: reinitialize metadata when new, opening, and deleting an AO

jenniferarnesen:
Implements [TECH-1049](https://dhis2.atlassian.net/browse/TECH-1049)

### Key features

1. Add CLEAR_METADATA to tSetInitMetadata and make sure it is called when use clicks New, Open and Delete. This will ensure that the metadata in the store doesn't grow in size and potentially become a performance bottleneck.

### DIFF
--- a/src/actions/metadata.js
+++ b/src/actions/metadata.js
@@ -24,7 +24,6 @@ export const tSetInitMetadata = () => (dispatch, getState) => {
         }
     })
 
-    console.log('tSetInitMetadata')
     dispatch(acClearMetadata())
     dispatch(acAddMetadata(metaData))
 }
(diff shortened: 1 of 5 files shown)


src/actions/metadata.js
```diff
@@ -1,12 +1,16 @@
...
     type: ADD_METADATA,
     value,
 })
 
+const acClearMetadata = () => ({
+    type: CLEAR_METADATA,
+})
+
 export const tSetInitMetadata = () => (dispatch, getState) => {
     const metaData = { ...getDefaultMetadata() }
```
martinkrulltott, line 15: if `acClearMetadata` adds `DEFAULT_METADATA`. doesn't this mean that we're adding default metadata twice then with the `...getDefaultMetadata`? 🤔

martinkrulltott (approved):

src/actions/metadata.js
jenniferarnesen: I see the confusion. CLEAR_METADATA sets the metadata to an empty object. I've renamed that constant in the reducer so it is hopefully more intuitive.

PR #63, opened 2022-03-25
fix: programStage issues when saving/loading AO

edoardo:
Implements [TECH-1059](https://dhis2.atlassian.net/browse/TECH-1059)

---

### Key features

1. avoid saving empty object for `programStage` in `current` TECH-1059
2. fix loading of AO when `programStage` is not present

---

### Description

`programStage` is not always selected in a AO.
Remove if empty when saving and don't assume it's present with an `id` when loading an AO.

---

### Screenshots

_supporting text_

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -251,8 +251,11 @@ const App = ({
             )
             const metadata = {
                 [program.id]: program,
-                [programStage.id]: programStage,
                 ...getDynamicTimeDimensionsMetadata(program, programStage),
+            }
+
+            if (programStage?.id) {
+                metadata[programStage.id] = programStage
             }
 
             addParentGraphMap(getParentGraphMapFromVisualization(visualization))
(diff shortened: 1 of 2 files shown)


martinkrulltott (approved):

src/components/App.js
```diff
@@ -251,10 +251,13 @@ const App = ({
             )
             const metadata = {
                 [program.id]: program,
-                [programStage.id]: programStage,
                 ...getDynamicTimeDimensionsMetadata(program, programStage),
             }
 
+            if (programStage?.id) {
```
martinkrulltott, line 257: if you want to keep it in the initial `metadata` assignment you could do `...(programStage?.id ? { [programStage.id]: programStage } : {})` above. But the separate assignment works too 👍

edoardo: I think it's more clear with a separate assignment.

PR #67, opened 2022-03-28
fix: various updates to duplicate dimensions (TECH-1032)

martinkrulltott:
Implements [TECH-1032](https://dhis2.atlassian.net/browse/TECH-1032)

---

### Key features

1. Shows program stage name in the data table headers when duplicates are present
2. Shows program stage name in the modal title for all data elements
3. Shows program stage name in the chip tooltip for all data elements
4. Prevents program stage name from getting stuck in the chip title when a duplicate is removed

---

### Screenshots

image
_program stage name in the data table headers for duplicates_

image
_program stage name in the modal title_

image
_program stage name in the chip tooltip for all data elements_

image
_program stage is hidden when a duplicate is removed_

### DIFF
--- a/src/components/Layout/DefaultLayout/DefaultAxis.js
+++ b/src/components/Layout/DefaultLayout/DefaultAxis.js
@@ -56,10 +56,10 @@ const DefaultAxis = ({ axisId, className }) => {
         const dimensions = dimensionIds.map((id) => {
             let dimension = {}
             if (metadata[id]) {
-                dimension = metadata[id]
+                dimension = { ...metadata[id] }
             } else {
                 const [rawDimensionId] = id.split('.').reverse()
-                dimension = metadata[rawDimensionId]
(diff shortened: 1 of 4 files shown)


src/components/Layout/DefaultLayout/DefaultAxis.js
```diff
@@ -56,10 +56,10 @@ const DefaultAxis = ({ axisId, className }) => {
         const dimensions = dimensionIds.map((id) => {
             let dimension = {}
             if (metadata[id]) {
-                dimension = metadata[id]
+                dimension = { ...metadata[id] }
```
jenniferarnesen, line 59: Is there any possibility that the metadata of the dimensions could have non-simple property types (object or array) that would therefore retain referential equality? I wasn't able to find any examples of that, so it looks safe. But just wanted to check with you on that.

PR #71, opened 2022-03-29
fix: interpretation modal scrolling

HendrikThePendric:
This addresses both the double scrollbar, and the data table being too high

### DIFF
--- a/src/components/Interpretations/InterpretationModal/InterpretationModal.js
+++ b/src/components/Interpretations/InterpretationModal/InterpretationModal.js
@@ -31,9 +31,6 @@ const modalCSS = css.resolve`
     aside.hidden {
         display: none;
     }
-    aside > :global(div) > :global(div) {
-        max-height: 100%;
-    }
 `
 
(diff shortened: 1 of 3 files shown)


src/components/Interpretations/InterpretationModal/InterpretationModal.js
```diff
@@ -22,16 +22,18 @@ import { useModalContentWidth } from './useModalContentWidth.js'
...
         width: auto !important;
         height: auto !important;
+        overflow-y: hidden;
     }
     aside.hidden {
         display: none;
     }
+    aside > :global(div) > :global(div) {
+        max-height: 100%;
+    }
```
HendrikThePendric, line 36: This part can be removed when https://github.com/dhis2/ui/pull/1013 is merged and we upgrade @dhis2/ui.

HendrikThePendric: This was done now 👍

PR #72, opened 2022-03-29
refactor: prevent unnecessary renders by not creating new empty objects

jenniferarnesen:
Two selectors were updated to return a constant object/array that represents the empty state if no value is found

sGetUiItemsByDimension
sGetUiConditionsByDimension

Chip.js: minor code cleanup
DimensionItemBase: removed default props for props that were previously removed (onClick doesn't need a default in this case)


janhenrikoverland (approved):
LGTM

PR #74, opened 2022-03-30
fix: avoid crash when opening tooltip or dimension modal in cases where option metadata is missing

jenniferarnesen:
Applies frontend fixes for [TECH-1056](https://dhis2.atlassian.net/browse/TECH-1056)

---

### Key features

1. For tooltip, if at least one of the option/legend names is not available, then don't add any html to the tooltip
image

3. For the dimension modal, if the name is not available, then use the code (i.e., Z4_VACCINE is the code in the screenshot below).
image

These fixes are workarounds and a proper backend fix should be coming in v2.39


jenniferarnesen:
Maybe we can work with this after all.

src/components/Layout/TooltipContent.js
```diff
@@ -85,7 +85,10 @@ export const TooltipContent = ({
         return itemDisplayNames
     }
 
-    const renderItems = (itemDisplayNames) => {
+    const renderItems = (itemDisplayNames = []) => {
+        if (itemDisplayNames.findIndex((name) => name === undefined) !== -1) {
```
martinkrulltott, line 89:
```suggestion
        if (!itemDisplayNames.find((name) => name === undefined)) {
```

martinkrulltott: A bit hard to follow what you're trying to do here though... do you want to make sure that all items to have a `name`?

martinkrulltott: if so then I think this is more clear
```suggestion
        if (itemDisplayNames.some((name) => name === undefined)) {
```

martinkrulltott: or even 
```suggestion
        if (itemDisplayNames.some((name) => !name)) {
```

jenniferarnesen: Yeah that looks better

src/components/Layout/TooltipContent.js
```diff
@@ -85,7 +85,10 @@ export const TooltipContent = ({
         return itemDisplayNames
     }
 
-    const renderItems = (itemDisplayNames) => {
+    const renderItems = (itemDisplayNames = []) => {
+        if (itemDisplayNames.some((name) => name === undefined)) {
```
martinkrulltott, line 89:
```suggestion
        if (itemDisplayNames.every((name) => name)) {
```

src/components/Layout/TooltipContent.js
```diff
@@ -85,7 +85,10 @@ export const TooltipContent = ({
         return itemDisplayNames
     }
 
-    const renderItems = (itemDisplayNames) => {
+    const renderItems = (itemDisplayNames = []) => {
+        if (itemDisplayNames.every((name) => name)) {
```
martinkrulltott, line 89:
```suggestion
        if (!itemDisplayNames.every((name) => name)) {
```

jenniferarnesen: I actually don't want this because if even one of the items is missing the name, then we don't want to show any of them.

martinkrulltott: yeah so "all need to have a name", right. So that means "some dont have name" or "not every have name"... now it says "every has name", which is the inverse of what you want, because I was a bit too fast on the keyboard and forgot the "!", whoopsi..

martinkrulltott (approved):

PR #75, opened 2022-04-01
fix: repetition index issues (TECH-1066)

martinkrulltott:
Implements [TECH-1066](https://dhis2.atlassian.net/browse/TECH-1066)

---

### Key features

1. Makes sure repetition is only saved when the value is non-default (TECH-1066)
2. Refactors parse and concat of dimensionId/programStageId to use util functions `formatDimensionId` and
`extractDimensionParts`
3. (BONUS) changes the prop type for `isInLayout` which was throwing console errors

### DIFF
--- a/src/modules/utils.js
+++ b/src/modules/utils.js
@@ -29,3 +29,15 @@ export const useDidUpdateEffect = (fn, inputs) => {
 
 export const formatDimensionId = (dimensionId, programStageId) =>
     programStageId ? `${programStageId}.${dimensionId}` : dimensionId
+
+export const extractDimensionParts = (input) => {
+    const [dimensionId, rawStageId] = input.split('.').reverse()
+    const [programStageId, repetitionIndex] = (rawStageId || '').split('[')
+    return {
+        dimensionId,
+        programStageId,
+        repetitionIndex:
+            repetitionIndex?.length &&
+            repetitionIndex.substring(0, repetitionIndex.indexOf(']')),
+    }
+}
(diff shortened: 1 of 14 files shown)


edoardo (approved):

src/modules/utils.js
```diff
@@ -29,3 +29,15 @@ export const useDidUpdateEffect = (fn, inputs) => {
 
 export const formatDimensionId = (dimensionId, programStageId) =>
     programStageId ? `${programStageId}.${dimensionId}` : dimensionId
+
+export const extractDimensionParts = (input) => {
```
jenniferarnesen, line 33: Could you add some unit tests for the function? That would also help to document what it does.

Also thinking that the function name should indicate that it is operating on the dimensionId, not the dimension object (which contains name and other properties).  Perhaps getDimensionIdParts, or extractDimensionIdParts

src/reducers/ui.js
```diff
@@ -337,6 +339,17 @@ export default (state = EMPTY_UI, action) => {
                 },
             }
         }
+        case REMOVE_UI_REPETITION: {
+            const {
+                [action.value]: removedProperty,
```
jenniferarnesen, line 344: I only find one instance of "removedProperty" in the code base. Is that expected?

jenniferarnesen: nvm, sorry this is just renaming...

src/modules/utils.js
martinkrulltott: Renamed + added unit tests 👍

jenniferarnesen (approved):

PR #82, opened 2022-04-05
fix: clear current in redux store when ui config is invalid and allow saving of AO drafts containing only program id.

jenniferarnesen:
Implements [TECH-1069](https://dhis2.atlassian.net/browse/TECH-1069)

### Key features

1. Clear current when validation on ui fails (when user clicks "Update" button). Validation fails if layout is missing or exists but is missing program id. This is less restrictive than before.
2. Saving an AO with only program id is now allowed so that user can create incomplete drafts.
3. Refactor: Convert to useSelector, useDispatch
4. Refactor: Give the `current.legacy` an explanatory function name
5. Cleanup: Remove commented-out code. This was from DV, which does not open a separate modal to show the AO version tied to the selected interpretation. Since LL opens a separate modal showing the vis tied to the interpretation, there is no navigating routes, and the Update button isn't even available.

https://user-images.githubusercontent.com/6113918/161753370-88f8df28-bf9d-451f-adce-6f449ee29575.mov


src/components/UpdateButton/UpdateVisualizationContainer.js
```diff
@@ -8,57 +8,27 @@ import {
...
-    onLoadingStart,
-    hideLayoutPanel,
-}) => {
     const onClick = () => {
         try {
-            validateLayout(getCurrentFromUi())
-            clearLoadError()
-            onUpdate()
-            onLoadingStart()
+            validateLayout(currentFromUi)
```
martinkrulltott, line 19: why not use the selector directly?
```suggestion
            validateLayout(useSelector(sGetCurrentFromUi))
```

jenniferarnesen: because I didn't think about it.

src/modules/layoutValidation.js
```diff
@@ -68,3 +68,16 @@ export const validateLayout = (layout) => {
...
 }
+
+export const layoutHasProgramId = (layout) => {
+    if (!layout) {
+        return false
+    }
+    switch (layout.type) {
+        case VIS_TYPE_LINE_LIST:
+        default:
+            return Boolean(layout?.program?.id)
```
jenniferarnesen, line 79: I guess the first `?` isn't necessary since line 73 checks that it exists.

src/modules/layoutValidation.js
```diff
@@ -68,3 +68,16 @@ export const validateLayout = (layout) => {
...
+    }
+}
+
+export const AOIsLegacy = (layout) => layout.legacy // The legacy prop is returned by the backend for AOs that were created in the old ER app
```
jenniferarnesen, line 83: bikeshedding here. Could the function be renamed to `aoCreatedInEventReportsApp`?

src/components/UpdateButton/UpdateVisualizationContainer.js
```diff
@@ -1,64 +1,40 @@
...
-import { sGetCurrent, sGetCurrentFromUi } from '../../reducers/current.js'
+import {
+    layoutHasProgramId,
+    validateLayout,
+} from '../../modules/layoutValidation.js'
+import { sGetCurrentFromUi } from '../../reducers/current.js'
+
+const UpdateVisualizationContainer = ({ renderComponent }) => {
+    const dispatch = useDispatch()
+    const currentFromUi = useSelector(sGetCurrentFromUi)
```
jenniferarnesen, line 18: You had commented previously to just use `useSelector(sGetCurrentFromUi)` directly instead of creating this variable. (I assume this can be done within a try/catch block). Did you still want to do that?

martinkrulltott: That's right, previously it was only use in one place so the `const` wasn't needed, but as its now being used in two places instead of only one I think it makes sense to assign it to a `const`

src/modules/layoutValidation.js
martinkrulltott, line 79:
```suggestion
            return Boolean(layout.program?.id)
```

src/modules/layoutValidation.js
martinkrulltott: renamed ✔️

edoardo (approved):

PR #87, opened 2022-04-08
fix: various interpretations fixes

edoardo:
Implements [TECH-1087](https://dhis2.atlassian.net/browse/TECH-1087), [TECH-1088](https://dhis2.atlassian.net/browse/TECH-1088).

---

### Key features

1. move interpretation form before the interpretations list
2. remove the onClick handler on the interpretation container
3. fix avatar initials, name, date vertical alignment

---

### Description

Since we have a big "See interpretation" button now instead of a small eye icon, there is no need to have the whole interpretation box clickable in order to open the interpretation view (modal).
This removes also the weird issue about the click event propagation which caused the interpretation view (modal) to open when interacting with the `SharingDialog`.

---

### Screenshots

Click handler removed:
(the cursor is not visible in the screenshot but it's now possible to select text)
Screenshot 2022-04-08 at 14 55 28

Interpretation form moved on top of the interpretations list:
Screenshot 2022-04-08 at 14 54 04

Vertical alignment:
before:
image (9)
after:
image (10)


src/components/Interpretations/common/Message/Message.js
```diff
@@ -25,21 +25,19 @@ const Message = ({ children, text, created, id, username, onClick }) => (
...
-                cursor: pointer;
             }
 
             .header {
                 display: flex;
                 gap: 6px;
                 align-items: center;
                 font-size: 13px;
-                line-height: 15px;
+                line-height: ${spacers.dp16};
```
martinkrulltott, line 35: I think we've concluded that we should only use `spacers` for actual spacing (margins, padding, gaps etc). So in this case and in the `font-size: ${spacers.dp12};` case below, we should use regular values instead (`16px`, `12px` respectively).

martinkrulltott, line 35:
```suggestion
                line-height: 16px;
```

src/components/Interpretations/common/Message/Message.js
```diff
@@ -25,21 +25,19 @@ const Message = ({ children, text, created, id, username, onClick }) => (
...
                 color: ${colors.grey900};
             }
 
             .header time {
                 font-size: ${spacers.dp12};
```
martinkrulltott, line 40:
```suggestion
                font-size: 12px;
```

src/components/Interpretations/common/Message/Message.js
edoardo: Fixed.

martinkrulltott (approved):

PR #91, opened 2022-04-11
fix: mouse pointer and clickable area for chip (TECH-1083)

martinkrulltott:
Implements [TECH-1083](https://dhis2.atlassian.net/browse/TECH-1083)

---

### Key features

1. Tweaks the css for chip to fix the mouse pointer on hover and clickable area

---

### Description

Moves the padding and margin from the container and context menu to the chip base, to make sure the hoverable and clickable areas fill the whole container, rather than having dead space in chip (since the chip base has the `onClick` etc)

Visually, the before and after are pixel perfect identical (in Mac Chrome).

---

### Screenshots

_before_
image

_after_
image

### DIFF
--- a/src/components/Layout/styles/Chip.module.css
+++ b/src/components/Layout/styles/Chip.module.css
@@ -15,7 +15,6 @@
     user-select: none;
     width: fit-content;
     align-items: center;
-    padding: 0 2px 0 6px;
 }
 
 .chipBase {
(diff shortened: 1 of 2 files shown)


src/components/Layout/styles/Chip.module.css
```diff
@@ -24,8 +23,10 @@
     max-width: 360px;
     align-items: center;
     font-size: 14px;
-    line-height: 17px;
+    line-height: 16px;
     color: var(--colors-grey900);
+    margin-right: var(--spacers-dp4);
+    padding: 4px 2px 4px 6px;
```
edoardo, line 29: Use spacers vars?

edoardo (approved):

PR #101, opened 2022-04-19
fix: format date and datetime

martinkrulltott:
### Key features

1. adds `max` to `datetime` and `date` condition inputs
2. formats the tooltip for `datetime` and `time`

---

### Screenshots

_set max to `date`_

image

_set max to `datetime`_

image

---

_`datetime` tooltip before_

image

_`datetime` tooltip after_

image

---

_`time` tooltip before_

image

_`time` tooltip after_

image

### DIFF
--- a/src/modules/conditions.js
+++ b/src/modules/conditions.js
@@ -83,6 +83,11 @@ export const VALUE_TYPE_TIME = 'TIME'
 export const VALUE_TYPE_DATETIME = 'DATETIME'
 export const VALUE_TYPE_ORGANISATION_UNIT = 'ORGANISATION_UNIT'
 
+export const API_TIME_DIVIDER = '.'
+export const UI_TIME_DIVIDER = ':'
+export const API_DATETIME_DIVIDER = 'T'
+export const UI_DATETIME_DIVIDER = ' '
+
 export const prefixOperator = (operator, isCaseSensitive) => {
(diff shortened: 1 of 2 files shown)


src/modules/conditions.js
```diff
@@ -220,6 +225,15 @@ export const getConditions = ({
             value = parts[1]
         }
 
+        if (
+            [VALUE_TYPE_TIME, VALUE_TYPE_DATETIME].includes(dimension.valueType)
+        ) {
+            value = value.replaceAll(API_TIME_DIVIDER, UI_TIME_DIVIDER)
```
edoardo, line 231: Should this be done with some kind of date/time formatting library instead?

edoardo (approved):

src/modules/conditions.js
edoardo: Assuming there is no edge-case with the backend format where the replace does not do the right thing, and assuming we want a fixed format regardless of user preference and locale, then this solution works fine.

martinkrulltott: Actually we figured out that the `datetime` conditions doesn't work, https://jira.dhis2.org/browse/DHIS2-13114, so I assume we'll have to adjust this at some point once the backend has a new format, but that'll be a separate PR later.

PR #105, opened 2022-04-21
fix: use different color on hover for action buttons TECH-1086

edoardo:
Implements [TECH-1086](https://dhis2.atlassian.net/browse/TECH-1086)

---

### Key features

1. use a different color when hovering the interpretation's action button

---

### Description

On hover a tooltip is shown but we also want to make it more visible that those are buttons.

---

### Screenshots
Screenshot 2022-04-21 at 16 30 54

Screenshot 2022-04-21 at 16 30 34

### DIFF
--- a/src/components/Interpretations/common/Message/MessageIconButton.js
+++ b/src/components/Interpretations/common/Message/MessageIconButton.js
@@ -28,7 +28,7 @@ const MessageIconButton = ({
                     disabled={disabled}
                 >
                     {count && count}
-                    <Icon color={selected ? colors.teal500 : colors.grey700} />
+                    <Icon className={cx('icon', { selected })} />
                 </button>
                 <style jsx>{`
                     .wrapper {
@@ -50,6 +50,20 @@ const MessageIconButton = ({
                     .button.selected {
                         color: ${colors.teal600};
                         font-weight: 500;
+                    }
+                    .button:hover {
+                        color: ${colors.grey900};
+                    }
+                    .icon {
+                        color: ${colors.grey700};
+                    }
+
+                    .icon.selected {
+                        color: ${colors.teal500};
+                    }
+
+                    .icon:hover {
+                        color: #000;
                     }
                 `}</style>
             </span>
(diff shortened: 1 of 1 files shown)


src/components/Interpretations/common/Message/MessageIconButton.js
```diff
@@ -51,6 +51,20 @@ const MessageIconButton = ({
                         color: ${colors.teal600};
                         font-weight: 500;
                     }
+                    .button:hover {
```
jenniferarnesen, line 54: Wondering if the selected button should have a stronger selected color (teal) rather than grey?

src/components/Interpretations/common/Message/MessageIconButton.js
```diff
@@ -51,6 +51,20 @@ const MessageIconButton = ({
...
+                        color: ${colors.grey900};
+                    }
+                    .icon {
```
jenniferarnesen, line 57: Hmm, I can't see that any html elements have the icon property.

edoardo: 🤔  good observation, I don't understand how come it works then...

src/components/Interpretations/common/Message/MessageIconButton.js
edoardo: Changed the hover color to a shade of teal

src/components/Interpretations/common/Message/MessageIconButton.js
```diff
@@ -51,6 +51,22 @@ const MessageIconButton = ({
...
+                    .button.selected:hover {
+                        color: ${colors.teal800};
+                    }
+
+                    .button.selected :global(svg) {
+                        color: ${colors.teal500};
+                    }
+
+                    .button.selected:hover :global(svg) {
+                        color: ${colors.teal700};
```
jenniferarnesen, line 68: The svg will be `teal700` and the button (line 60) will be `teal800`. Are these intended to be different?

edoardo: Because the SVG has such thin lines, `teal800` almost looked grey, so I went for a lighter version.

jenniferarnesen (approved):

PR #112, opened 2022-04-27
fix: preserve fetched dimensions when collapsing Program Dimensions panel

jenniferarnesen:
Implements [TECH-1089](https://dhis2.atlassian.net/browse/TECH-1089)

---

### Key features

1. In order to preserve the fetched dimensions, ProgramDimensionsList cannot be un-mounted when collapsing the Program Dimensions panel, so that the useProgramDimensions hook retains its local state.
2. Very small refactor: move the `userSettings` hook to useProgramDimensions, where it is actually needed.

---

### Known issues

-   When expanding the panel, the list will scroll up to the top. Not sure if this is an annoyance or a bonus.
-  There might be a teeny-tiny bit of sluggishness when expanding the panel, if the dimension list is very long (due to lots of previous scrolling down)

---

### Screenshots

https://user-images.githubusercontent.com/6113918/165491134-265bec51-b375-4ffb-bc32-420edb4f1feb.mov

### DIFF
--- a/src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
+++ b/src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
@@ -104,7 +104,7 @@ const ProgramDimensionsPanel = ({ visible }) => {
         setDimensionType(DIMENSION_TYPE_ALL)
     }, [inputType, selectedProgramId])
 
-    if (!visible || !called) {
+    if (!called) {
         return null
     }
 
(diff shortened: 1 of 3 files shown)


src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
```diff
@@ -131,46 +131,51 @@ const ProgramDimensionsPanel = ({ visible }) => {
...
-                              )
-                            : i18n.t(
-                                  'Choose a program above to add program dimensions.'
-                              )}
-                    </div>
-                )}
-            </div>
+                </div>
+            )}
+            {visible && (
```
edoardo, line 144: Could this div be part of the above condition?

edoardo (approved):

src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsPanel.js
jenniferarnesen: yep

PR #121, opened 2022-04-29
fix: don't rely on the app's Redux store for header names TECH-1144

edoardo:
Partially implements [TECH-1144](https://dhis2.atlassian.net/browse/TECH-1144)

**Requires https://github.com/dhis2/analytics/pull/1282**

---

### Key features

1. use analytics header data for columns names instead of reading app's Redux store
2. remove the need for `useAvailableWidth` hook to set a good value for `scrollWidth` in the `DataTable`

---

### Description

This component eventually will become the LL plugin so it cannot rely on the app's Redux store.
All it needs must either be defined within the component, passed as props, or retrieved from the analytics response.
For the column headers we looked into the app's Redux store metadata, where columns names for time dimensions are read from the program data.

The backend should return the correct name for the header based on the current program.
Once the backend fix is ready, the "aliased" headers will have the correct name to display in the table's column headers.

About the `useAvailableWidth` hook, the `DataTable` scrolling issues are now solved with CSS and there is no need to compute a value for `scrollWidth` which is now always `100%`.


martinkrulltott (approved):

HendrikThePendric (approved):
👍 This has really cleaned things up. Well done.

PR #122, opened 2022-05-02
fix: do not render ProgramDimensionsList if panel is closed

jenniferarnesen:
Implements [TECH-1136](https://dhis2.atlassian.net/browse/TECH-1136)
Reverts (minus the refactor) of https://github.com/dhis2/line-listing-app/pull/112
---

### Key features

1. Turns out that expanding/collapsing the ProgramDimensionsPanel becomes sluggish when several pages of program dimensions have been loaded, especially when cpu is throttled. The sluggishness is less marked with the original solution, which is to refetch the dimensions when expanding.


martinkrulltott (approved):
LGTM. `en.pot` updates?

PR #143, opened 2022-05-11
ci: replace semantic commit checks with new workflow

ismay:


jenniferarnesen (approved):

src/components/MainSidebar/InputPanel/InputPanel.js
```diff
@@ -23,12 +23,13 @@ const getLabelForInputType = (type) => {
 }
 
 const InputPanel = ({ visible }) => {
+    const selectedInput = useSelector(sGetUiInput)
+    const dispatch = useDispatch()
+
     if (!visible) {
         return null
     }
```
ismay, line 31: @HendrikThePendric This looks like a good fix to me. You ok with this as well?

PR #148, opened 2022-05-16
chore: unit tests for conditions module

jenniferarnesen:
Tests for upcoming changes related to [TECH-1150](https://dhis2.atlassian.net/browse/TECH-1150)

---

### Key features

1. unit tests for conditions

Question: we do not differentiate between case sensitive and non-case sensitive in the tooltip. Intentional, right?

### DIFF
--- a/src/modules/__tests__/conditions.spec.js
+++ b/src/modules/__tests__/conditions.spec.js
@@ -1,0 +1,160 @@
+import { getConditions } from '../conditions.js'
+
+test('Legend set chosen with no conditions selected', () => {
+    const conditions = {
+        condition: '',
+        legendSet: 'legendSetId1',
+    }
+    const metadata = {
+        legendSetId1: { id: 'legendSetId1', name: 'Legend Set Name' },
+    }
+    const dimension = {}
+    const actual = getConditions({ conditions, metadata, dimension })
+
+    expect(actual).toEqual(['Legend Set Name'])
+})
+
+test('Legend set chosen with conditions selected', () => {
+    const conditions = {
+        condition: 'IN:Condition1Id;Condition2Id',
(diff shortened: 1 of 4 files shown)


src/modules/__tests__/conditions.spec.js
```diff
@@ -0,0 +1,170 @@
...
+    const metadata = {
+        legendSetId1: { id: 'legendSetId1', name: 'Legend Set Name' },
+    }
+    const dimension = {}
+    const actual = getConditions({ conditions, metadata, dimension })
+
+    expect(actual).toEqual(['Legend Set Name'])
+})
+
+test('Legend set chosen with conditions selected', () => {
```
martinkrulltott, line 17:
```suggestion
test('Legend set chosen with legends selected', () => {
```

src/modules/__tests__/conditions.spec.js
```diff
@@ -0,0 +1,170 @@
...
+    {
+        dimensionValueType: 'INTEGER',
+        condition: 'GT:3568',
+        expected: ['Greater than (>): 3568'],
+    },
+    {
+        dimensionValueType: 'LONG_TEXT',
+        condition: 'ILIKE:Cats',
+        expected: ['Contains: Cats'],
+    },
```
martinkrulltott, line 145: Add a case for `!ILIKE` as well?

src/modules/__tests__/conditions.spec.js
martinkrulltott: Also change `{ id: 'Condition1Id', name: 'Condition 1' }` to ``{ id: 'Legend1Id', name: 'Legend 1' }`, as to not confuse conditions with legends.

martinkrulltott (approved):

PR #154, opened 2022-05-20
feat: legend set table coloring (DHIS2-75)

martinkrulltott:
Implements [DHIS2-75](https://jira.dhis2.org/browse/DHIS2-75)

**Requires https://github.com/dhis2/analytics/pull/1249**

---

### Key features

1. Enable the legend tab in the options modal
2. Apply legend colors to the table
3. Conditionally display the legend key to the right of the table

---

### Description

Adds support for using legends in the data table by enabling it in the options modal. Inspired by the similar implementation of the legend options tab, legend colors and legend key for pivot tables in DV.

Minor refactoring along the way:
- Completely removes `acSetUiOption`
- Refactors `VALUE_TYPE_...` to be fetched from Analytics (instead of having duplicates of these in the app as well)
- Refactors some strings to be consts, e.g. `OPTION_DISPLAY_DENSITY` and `SEPARATOR_NONE`

---

### Known issues

-   [x] Background color still WIP (https://github.com/dhis2/ui/pull/1069)
-   [x] CSS issue with the legend key with wide tables

---

### Screenshots

_overview_
image

_options modal_
image

_legend in table_
image

_setting the background color_
image

_legend key_
image

_css issue with wide tables - a part of the legend key gets pushed outside of the window_
image


src/modules/current.js
```diff
@@ -37,7 +43,21 @@ export const getProgramStageFromUi = (ui) =>
           }
         : {}
 
-export const getOptionsFromUi = (ui) => pick(ui.options, Object.keys(options))
+export const getOptionsFromUi = (ui) => {
+    const savableOptions = pick(
+        ui.options,
+        Object.keys(options).filter((option) => options[option].savable)
```
edoardo, line 49: This is a "breaking change" for this function, as it does not return the same set of options.
I'm not sure if we can change it like this.
It might be fine for LL, but the idea (at least in DV) I think is that all options go into `current` which is passed to the plugin and when saving, the `savable` options are the only ones that end up in the object POSTed to the endpoint.

edoardo: Same thing below.

edoardo (approved):

PR #163, opened 2022-06-07
chore: add more cypress tests and extract some common functions

jenniferarnesen:
Implements [DHIS2-13375](https://jira.dhis2.org/browse/DHIS2-13375)

---

### Key features

1. new tests added for options, numeric and boolean conditions, and creating an event and an enrollment
2. added data-test attributes when it was not reasonable otherwise to select the correct element
3. removed data-test attributes not needed or in use (they can judiciously be added later when tests that need it are added)
4. a few data-test attributes updated
5. tests now run against live instance on ci
6. a couple of useful functions extracted to a /helpers directory for sharing. These can be tweaked and expanded later as needed. Trying not to over-engineer too much.
7. downgrade cypress to v8 since the tests were failing on CI with v9. Also v8 is the peer dependency version for the dhis2-utils-cypress and cypress-commands.


cypress/integration/create_enrollment.cy.js
```diff
@@ -0,0 +1,129 @@
...
+const periodLabel = 'Date of enrollment'
+
+const setUpTable = () => {
+    selectProgramDimensions({
+        inputType: INPUT_ENROLLMENT,
+        programName: 'Child Programme',
+        dimensions: [dimensionName],
+    })
+
+    selectPeriod({
```
martinkrulltott, line 20: We're going to be doing this every time we create a new AO, which is a lot. I see the need for a one-liner function call to do this, instead of having to repeat the `periodLabel` and all the props everywhere, which would mean a lot of copy-pasting and a big risk for typos along the way. I suggest we add `dataTest` to the periods so we can click on the regardless of what they're called and just call something like `selectPeriodLast12Months` or `selectRelativePeriod(LAST_12_MONTHS)`.

cypress/integration/options.cy.js
```diff
@@ -0,0 +1,110 @@
...
+    selectPeriod,
+    FIXED,
+    getPreviousYearStr,
+    getLineListTable,
+} from '../helpers/index.js'
+import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+describe('options', () => {
+    it('sets comfortable display density', () => {
+        cy.visit('#/R4wAb2yMLik', EXTENDED_TIMEOUT)
```
martinkrulltott, line 13: Just a note, in addition to making sure that old AOs work we should also test the complete flow for setting these options from scratch. Might be good to add a `// TODO: ` comment as a reminder.

cypress/integration/options.cy.js
```diff
@@ -0,0 +1,110 @@
...
+        cy.contains('Comma').click()
+        cy.getWithDataTest('{options-modal-actions}').contains('Update').click()
+
+        getLineListTable()
+            .find('tbody > tr')
+            .eq(0)
+            .find('td')
+            .eq(1)
+            .should('contain', '1,232')
+    })
```
martinkrulltott, line 109: Add a test for setting it back to `None` and asserting that the value is `1232`?

cypress/integration/conditions/boolean_conditions.cy.js
```diff
@@ -0,0 +1,168 @@
...
+    )
+}
+
+describe('boolean conditions', () => {
+    beforeEach(() => {
+        cy.visit('/', EXTENDED_TIMEOUT)
+        setUpTable()
+    })
+
+    it('Yes only', () => {
```
martinkrulltott, line 70: Maybe we can rename this just to not confuse it with the `TRUE_ONLY` value type (my head just slipped on this for a few seconds 🙈 . e.g. `Yes selected` or `Only yes selected`...
```suggestion
    it('Yes selected', () => {
```

cypress/integration/conditions/numeric_conditions.cy.js
```diff
@@ -0,0 +1,286 @@
...
+        assertTooltipContainsEntries([
+            'Program stage: Birth',
+            'Less than or equal to (≤): 2 615',
+        ])
+    })
+
+    it('not equal to', () => {
+        addConditions([{ conditionName: 'not equal to', value: '1232' }])
+
+        assertTableMatchesExpectedRows([
```
martinkrulltott, line 152: Could we maybe find test data with less expected rows? 20+ rows feels tricky to debug if one happens to be missing for some reason in the future... I suggest something in the range of 3-7 rows should be enough to test it properly but few enough to be able to debug easily.

martinkrulltott (approved):

cypress/integration/create_enrollment.cy.js
martinkrulltott: I've refactored this to use a list of `TEST_RELATIVE_PERIODS` and `TEST_FIXED_PERIODS` instead. However maybe its best to just assign these objects proper names instead (like `LAST_12_MONTHS` above) so its more straight-forward to see what's actually being used (rather than just e.g. `TEST_RELATIVE_PERIODS[0]` which has no semantic meaning.)

cypress/integration/options.cy.js
martinkrulltott: My bad, I thought the opened AO contained set options already, but I see now that they are just plain AOs (since any AO can use these options anyway), so disregard my comment above.

jenniferarnesen:
Changes LGTM

PR #167, opened 2022-06-10
fix: use feature toggling for the legend tab in options (DHIS2-75)

martinkrulltott:
Implements [DHIS2-75](https://jira.dhis2.org/browse/DHIS2-75)

---

### Key features

1. Prevent the legend tab from being displayed on servers < 2.38.1

---

### Description

This is based on the theory (currently untested) that only the saving part for legend sets needs to be feature toggled. As legend sets themselves are optional, so if the user can't save an AO with one then legend sets will never be present when loading (since the `/eventVisualizations` endpoint handles non-existing params gracefully).

### DIFF
--- a/src/modules/options/lineListConfig.js
+++ b/src/modules/options/lineListConfig.js
@@ -17,8 +17,9 @@ export default (serverVersion) => [
             ]),
         },
     ]),
-    ...(serverVersion.minor >= 39 ||
-    (serverVersion.minor === 38 && serverVersion.patch >= 1)
+    ...(`${serverVersion.major}.${serverVersion.minor}.${
+        serverVersion.patch || 0
+    }` > '2.38.1'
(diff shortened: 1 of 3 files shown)


src/modules/options/lineListConfig.js
```diff
@@ -17,5 +17,9 @@ export default () => [
             ]),
         },
     ]),
-    getLegendTab(),
+    ...(`${serverVersion.major}.${serverVersion.minor}.${
+        serverVersion.patch || 0
+    }` > '2.38.1'
```
edoardo, line 22: This doesn't work in some cases, for example `'12.0.0' > '2.38.1'`, probably not a problem in the short term, but if we ever decide to skip to a major `>=10` we need to change the comparison.

edoardo (approved):

PR #168, opened 2022-06-14
fix: untie Plugin component from the app's Redux store

edoardo:
### Key features

1. replace `acSetLoadError` with a `onError` callback
2. replace `acSetUiDimensionModal` with a `onColumnHeaderClick` callback

---

### Description

The plugin cannot rely on the app's Redux store, as it should be standalone so it can be reused in different places, like dashboard app.

The error handling is now using a callback, similarly to how it's done already in the DV plugin.
If there is an error in the plugin (most likely an error returned from the analytics endpoint) and if the `onError` callback prop is provided, the plugin simply calls the callback.
The error is now handled in the `App` component.

In the LL app we want the column header to open the correct dimension modal when clicked, this is now also achieved via a callback, so the app is responsible of setting the modal id in its Redux store.

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -24,10 +24,7 @@ import cx from 'classnames'
 import moment from 'moment'
 import PropTypes from 'prop-types'
 import React, { useState, useEffect } from 'react'
-import { useDispatch, useSelector } from 'react-redux'
-// TODO this action cannot be used and a callback prop should be passed instead
-// when the plugin is used in dashboard, this feature should be disabled
-import { acSetUiOpenDimensionModal } from '../../actions/ui.js'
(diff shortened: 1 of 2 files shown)


src/components/Visualization/Visualization.js
```diff
@@ -133,26 +123,8 @@ export const Visualization = ({
...
-                    break
-                default:
-                    output = genericServerError()
-            }
-        } else {
-            output = genericServerError()
-        }
-        dispatch(acSetLoadError(output))
+    if (error && onError) {
+        onError(error)
```
janhenrikoverland, line 127: Could be error && onError && onError(error) on a single line, but keep it if you prefer the readability.

janhenrikoverland (approved):

PR #175, opened 2022-08-05
test: refactor lists to individual objects

martinkrulltott:
Using lists for test data started as a good idea but turned out to be very confusing to use, e.g. `SOME_LIST[0]` gives no contextual meaning to which object it's referring to.
The lists are now split into individual objects which are easier to pick and understand.

### DIFF
--- a/cypress/helpers/dimensions.js
+++ b/cypress/helpers/dimensions.js
@@ -29,12 +29,18 @@ const selectProgramAndStage = ({ inputType, programName, stageName }) => {
 export const selectEventProgram = ({ programName, stageName }) =>
     selectProgramAndStage({ inputType: INPUT_EVENT, programName, stageName })
 
-// export const selectEnrollmentProgram = ({ programName, stageName }) =>
-//     selectProgramAndStage({
-//         inputType: INPUT_ENROLLMENT,
-//         programName,
-//         stageName,
-//     })
+export const selectEnrollmentProgram = ({ programName, stageName }) =>
+    selectProgramAndStage({
+        inputType: INPUT_ENROLLMENT,
+        programName,
+        stageName,
+    })
+
+export const openDimension = (dimensionName) => {
+    cy.getWithDataTest('{program-dimension-list}')
+        .contains(dimensionName)
+        .click()
+}
 
 const selectProgramDimensions = ({
     inputType,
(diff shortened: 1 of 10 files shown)


cypress/helpers/dimensions.js
```diff
@@ -29,12 +29,18 @@ const selectProgramAndStage = ({ inputType, programName, stageName }) => {
...
-//     })
+export const selectEnrollmentProgram = ({ programName, stageName }) =>
+    selectProgramAndStage({
+        inputType: INPUT_ENROLLMENT,
+        programName,
+        stageName,
+    })
+
+export const openDimension = (dimensionName) => {
+    cy.getWithDataTest('{program-dimension-list}')
```
jenniferarnesen, line 40: Can `getBySel` be used here or no?

cypress/integration/conditions/numeric_conditions.cy.js
```diff
@@ -11,17 +15,17 @@ import {
 } from '../../helpers/table.js'
 import { EXTENDED_TIMEOUT } from '../../support/util.js'
 
-const event = TEST_EVENT_DATA[1]
-const dimensionName = event.dimensions[0]
+const event = ANALYTICS_PROGRAM
```
jenniferarnesen, line 18: Is it needed to define new constants or can you just plug ANALYTICS_PROGRAM directly into where it is used (eg line 24)

jenniferarnesen (approved):

cypress/integration/conditions/numeric_conditions.cy.js
martinkrulltott: Good point, yeah I was also considering it. But I wasn't sure how dependent on the Analytics program we'd want to be. It's used in two places (see code below), so I figured a const was useful. But since everything is built on the Analytics program should maybe consider not even passing it around in the first place... Either way I'd leave that for a separate PR though.

```
const event = ANALYTICS_PROGRAM
const dimensionName = TEST_DIM_NUMBER
const periodLabel = event[DIMENSION_ID_EVENT_DATE]
const stageName = 'Stage 1 - Repeatable'

const setUpTable = () => {
    selectEventProgramDimensions({ ...event, dimensions: [dimensionName] })
```

PR #180, opened 2022-08-08
test: alphanumeric conditions + unsupported types + org unit

martinkrulltott:
- New tests for alphanumeric types
- New tests for unsupported types (age, coordinate)
- New tests for the org unit type

### DIFF
--- a/cypress/data/index.js
+++ b/cypress/data/index.js
@@ -22,23 +22,14 @@ export const TEST_AOS = [
     },
 ]
 
-// alpha-numerics
+// alphanumerics
 export const TEST_DIM_TEXT = 'Analytics - Text'
-// VALUE_TYPE_LETTER
-// VALUE_TYPE_LONG_TEXT
-// VALUE_TYPE_EMAIL
-// VALUE_TYPE_USERNAME
-// VALUE_TYPE_URL
-// VALUE_TYPE_PHONE_NUMBER
-
-// booleans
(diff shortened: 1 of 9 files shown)


cypress/integration/conditions/numeric_conditions.cy.js
```diff
@@ -185,6 +190,84 @@ describe('number conditions', () => {
...
+        expectTableToBeVisible()
+
+        cy.getBySelLike('layout-chip').contains(`${dimensionName}: all`)
+    })
+
+    it('set only', () => {
+        addPreset(TEST_PRESET)
+
+        expectTableToMatchRows([
+            '2021-11-01',
```
jenniferarnesen, line 243: Should `currentYear` or `previousYear` be used instead?

cypress/data/index.js
```diff
@@ -22,23 +22,14 @@ export const TEST_AOS = [
     },
 ]
 
-// alpha-numerics
+// alphanumerics
 export const TEST_DIM_TEXT = 'Analytics - Text'
```
jenniferarnesen, line 26: What do you think about dropping the `TEST_` prefix? They are the real dimensions. Instead they could be called `DIMENSION_TEXT` or `DIM_TEXT`. Technically these (until line 59) are data elements, though maybe we don't want to get that specific.

jenniferarnesen (approved):

PR #183, opened 2022-08-10
test: option set conditions

martinkrulltott:
- Add tests for option set conditions
- Add additional tests to all other conditions to make sure they display properly in the visualization (to avoid backend issues like the one found while working on option sets)
- en.pot updates from https://github.com/dhis2/line-listing-app/pull/179 (`-` unescapes the string to prevent / etc in the searchTerm to be shown as unicode chars. https://www.i18next.com/translation-function/interpolation#unescape)


cypress/integration/conditions/alphanumeric_conditions.cy.js
```diff
@@ -244,6 +245,23 @@ describe('alphanumeric types', () => {
                     operator
                 )
             })
+            cy.getBySel('alphanumeric-condition-type')
+                .closest('[data-test=dhis2-uicore-layer]')
```
jenniferarnesen, line 249: There are 4 places where this data-test id is used in this manner. Is it possible to make a custom chainable cy command to do this? e.g. `closeCoreLayer`

image

i18n/en.pot
```diff
@@ -192,8 +192,8 @@ msgstr "Showing all values for this dimension"
 msgid "Program stage: {{stageName}}"
 msgstr "Program stage: {{stageName}}"
 
-msgid "No dimensions found for '{{searchTerm}}'"
-msgstr "No dimensions found for '{{searchTerm}}'"
+msgid "No dimensions found for '{{- searchTerm}}'"
```
jenniferarnesen, line 195: What does the "-" do?

jenniferarnesen:
File renaming suggestions:
option_set_condition => optionset_conditions
orgunit_condition => orgunit_conditions

i18n/en.pot
martinkrulltott: Unescapes the string to prevent `/` etc in the searchTerm to be shown as unicode chars. https://www.i18next.com/translation-function/interpolation#unescape

cypress/integration/conditions/alphanumeric_conditions.cy.js
martinkrulltott: Either that or we could at least extract that into a `closePopper` function to only do it once. I'll refactor it in my next PR

i18n/en.pot
jenniferarnesen: Could you add that to the PR description (and final merge commit)

jenniferarnesen (approved):

PR #184, opened 2022-08-10
chore: add unit tests for axis and conditions modules

jenniferarnesen:
Add unit tests for modules/axis.js and modules/conditions.js
I've made some code comments regarding the relevant functions


src/modules/__tests__/conditions.spec.js
```diff
@@ -250,3 +255,144 @@ describe('conditions that do not include legend sets or option sets', () => {
...
+        {
+            operator: '!LIKE',
+            isCaseSensitive: false,
+            expected: '!ILIKE',
+        },
+        {
+            operator: 'LIKE',
+            isCaseSensitive: false,
+            expected: 'ILIKE',
+        },
```
martinkrulltott, line 328: Add a case for `!LIKE` -> `!LIKE` and `!EQ` -> `!EQ` (i.e. the "not" operator with `isCaseSensitive: true`)

src/modules/conditions.js
```diff
@@ -124,6 +126,10 @@ export const unprefixOperator = (operator) => {
     }
 }
 
+// TODO - in practice this function isn't used for the 'IN' operator
+// but if it were the result would be wrong. The function
+// should probably control for the allowed operators and throw if the
+// operator isn't one of the allowed ones.
```
martinkrulltott, line 132: Good point. You could add a check in the beginning of the function like 
`if(!['EQ','!EQ'.....].includes(operator) {throw 'some message...'}` and add the error to the `expected` prop for the `IN` case in the tests above.

martinkrulltott (approved):

src/modules/conditions.js
jenniferarnesen: I'll address this in a later PR since this increases the code and test scope somewhat (handling potential thrown errors).

PR #185, opened 2022-08-10
test: add tests for interpretations components

edoardo:
**Requires https://github.com/dhis2/analytics/pull/1305**

---

### Key features

1. tests for the interpretations components

---

### Description

Cypress tests for the new interpretation components. unit (in side panel) and modal.
It covers basic interaction for:
* opening/closing the panel
* adding an interpretation
* editing an interpretation
* viewing an interpretation in the modal
* changing the page size in the modal and verify that it doesn't change in the main view
* add a comment to the interpretation
* close the modal
* delete an interpretation


cypress/integration/interpretations.cy.js
```diff
@@ -0,0 +1,169 @@
...
+            TEST_INTERPRETATION_TEXT
+        )
+        cy.getBySel('interpretations-list').should(
+            'contain',
+            'See interpretation'
+        )
+    })
+
+    it('the new interpretation can be edited', () => {
+        cy.getBySel('interpretation-edit-button').first().click()
```
martinkrulltott, line 91: why `first`?

cypress/integration/interpretations.cy.js
```diff
@@ -0,0 +1,169 @@
...
+        )
+
+        // the page size can be changed
+        // (change page size for testing that the sorting does not apply to the main view)
+        cy.getBySel('interpretation-modal').contains('Rows per page').click()
+        cy.getBySel('dhis2-uicore-select-menu-menuwrapper')
+            .contains('5')
+            .click()
+        cy.getBySel('interpretation-modal')
+            .find('[data-test="dhis2-uiwidgets-pagination-pagesize-select"]')
```
martinkrulltott, line 146:
```suggestion
            .findBySel('dhis2-uiwidgets-pagination-pagesize-select')
```

cypress/integration/interpretations.cy.js
```diff
@@ -0,0 +1,169 @@
...
+            .should('contain', '5')
```
martinkrulltott, line 147:
```suggestion
            .contains('5')
```
This applies to more places in this file

martinkrulltott (commented):
wrote a few comments above.

generally, `should('contain', ...)` can be replaced by `contains(...)` throughout the file. that seems to be our best practice for our other tests

also, skip `TEST_AOS[0]` and just add an object with a unique name instead, like we have `HIV_PROGRAM`

cypress/integration/interpretations.cy.js
edoardo: Yes, but since this is an existing AO that we open, are we always 100% sure there isn't any other interpretation?

martinkrulltott: if you pick out an existing AO that only has 1 interpretation then there should only be 1 unless we explicitly change this in the db template, which we shouldn't. So I'd say it's actually a good thing if it fails because there are more than 1, since that means that something is obviously wrong.

martinkrulltott: the reason for using https://test.e2e.dhis2.org/lytics-apps-dev/ is that it should only be used by automated tests, which in turn should be self-cleaning. So the db can be expected to be exactly the same over time.

edoardo: Removed `first()` 👍

cypress/data/index.js
```diff
@@ -24,6 +24,11 @@ export const HIV_PROGRAM = {
     [DIMENSION_ID_LAST_UPDATED]: 'Last updated on',
 }
 
+export const TEST_AO = {
+    id: 'qmuoJ2ROwUb',
+    name: 'List of patients with no treatment initiation date',
+}
+
 export const TEST_AOS = [
```
martinkrulltott, line 32: remove `TEST_AOS` completely

cypress/data/index.js
```diff
@@ -24,6 +24,11 @@ export const HIV_PROGRAM = {
     [DIMENSION_ID_LAST_UPDATED]: 'Last updated on',
 }
 
+export const TEST_AO = {
```
martinkrulltott, line 27: rename to something that gives semantic meaning, e.g. `TEST_AO_WITH_INTERPRETATION` or something, to distinguish it from other test AOs that we'll add later for different use-cases..

martinkrulltott (approved):
LGTM, just a tiny thing with test AO that needs to be adjusted

cypress/data/index.js
edoardo: Removed and adjusted `smoke` and `options` test files.

cypress/data/index.js
edoardo: There's nothing special with the AO used for interpretations tests, so we stick to one `TEST_AO` for now.

martinkrulltott (approved):

PR #187, opened 2022-08-12
test: legend set (DHIS2-75)

martinkrulltott:
Tests for [DHIS2-75](https://dhis2.atlassian.net/browse/DHIS2-75)


cypress/helpers/table.js
```diff
@@ -1,5 +1,12 @@
 import { EXTENDED_TIMEOUT } from '../support/util.js'
 
+export const expectAOTitleToBeValue = (value) =>
```
jenniferarnesen, line 3: Since it is checking "contain", maybe the function title should be `expectAOTitleToContain`

cypress/integration/legendSet.cy.js
```diff
@@ -0,0 +1,313 @@
...
+    const TEST_CELLS = [
+        { value: 60, color: 'rgb(66, 146, 198)' },
+        { value: 35, color: 'rgb(158, 202, 225)' },
+    ]
+
+    const TEST_LEGEND_SET = { name: 'Alert', color: 'rgb(237, 227, 99)' }
+
+    const assertCellsHaveDefaultColors = (selector) =>
+        cy
+            .getBySel('table-body')
```
jenniferarnesen, line 50: Is it possible to use any of the helper functions in helpers/table.js ?

jenniferarnesen (approved):

cypress/integration/legendSet.cy.js
martinkrulltott: No, this is quite specific to legend sets. Also I didn't want to chain the `.each` to a proprietary function, so I thought I'd stick to vanilla for the whole line.

edoardo (approved):

PR #189, opened 2022-08-22
test: repeated events

martinkrulltott:
Adds tests for the repeated events tab in the conditions modal, which is only enabled for repeatable stages in enrollment programs.

### DIFF
--- a/cypress/integration/repeated_events.cy.js
+++ b/cypress/integration/repeated_events.cy.js
@@ -4,6 +4,7 @@ import {
     openDimension,
     selectEnrollmentProgram,
     selectEnrollmentProgramDimensions,
+    selectEventProgram,
 } from '../helpers/dimensions.js'
 import { clickMenubarUpdateButton } from '../helpers/menubar.js'
 import { selectRelativePeriod } from '../helpers/period.js'
@@ -13,6 +14,9 @@ import {
     getTableDataCells,
 } from '../helpers/table.js'
 import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+const getRepeatedEventsTab = () =>
+    cy.getBySel('conditions-modal-content').contains('Repeated events')
 
 const setUpTable = ({ enrollment, dimensionName }) => {
     selectEnrollmentProgramDimensions({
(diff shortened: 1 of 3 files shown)


cypress/integration/repeated_events.cy.js
```diff
@@ -0,0 +1,207 @@
...
+        expectHeaderToContainExact(
+            4,
+            'Analytics - Percentage - Stage 1 - Repeatable (most recent -1)'
+        )
+        expectHeaderToContainExact(
+            5,
+            'Analytics - Percentage - Stage 1 - Repeatable (most recent)'
+        )
+    })
+    it.skip('repetition out of bounds returns as a empty value', () => {
```
martinkrulltott, line 163: This is the backend issue we were discussing @janhenrikoverland

cypress/integration/repeated_events.cy.js
```diff
@@ -0,0 +1,207 @@
...
+            4,
+            'Analytics - Percentage - Stage 1 - Repeatable (most recent -1)'
+        )
+        expectHeaderToContainExact(
+            5,
+            'Analytics - Percentage - Stage 1 - Repeatable (most recent)'
+        )
+    })
+    it.skip('repetition out of bounds returns as empty value', () => {
+        // FIXME: Backend issue, the repetition out of bounds is returned as 0 when it should be an empty string
```
martinkrulltott, line 164: This is the backend issue we were discussing @janhenrikoverland

cypress/integration/repeated_events.cy.js
```diff
@@ -0,0 +1,207 @@
...
+        .should('be.visible')
+        .and('have.value', recent)
+    cy.getBySel('oldest-input')
+        .find('input')
+        .should('be.visible')
+        .and('have.value', oldest)
+    cy.getBySel('conditions-modal').contains('Hide').click()
+}
+
+const expectHeaderToContainExact = (index, value) =>
```
edoardo, line 77: Should this be in the table helper?

edoardo (approved):

cypress/integration/repeated_events.cy.js
martinkrulltott: It could be, but as its only used in this file (so far) and has a pretty specific use case (checking if the header at a specific index has an exact match) I think we could leave it in the file until/if we find another use case for it.

PR #191, opened 2022-08-23
test: translations with i18n

martinkrulltott:
Adds tests for asserting that translations from i18n display correctly.


janhenrikoverland (commented):
Approved when test is enabled.

janhenrikoverland (approved):

janhenrikoverland (approved):

PR #209, opened 2022-09-16
fix: click on column header opens dimension modal DHIS2-13738

edoardo:
Implements [DHIS2-13738](https://dhis2.atlassian.net/browse/DHIS2-13738)

---

### Key features

1. fix regression: click on column header did not open correspondent dimension modal

---

### Description

This fixes a regression that caused some header columns to not open the correspondent dimension modal when clicked.
The affected column were all the ones where the dimension name is remapped to the name passed to the analytics request in the header parameter.

---

### Screenshots

Click on Org Unit column opens Org Unit dimension dialog:
Screenshot 2022-09-16 at 10 06 15

Click on a time dimension column opens the correspondent dimension dialog:
Screenshot 2022-09-16 at 10 06 05

### DIFF
--- a/cypress/data/index.js
+++ b/cypress/data/index.js
@@ -4,7 +4,6 @@ import {
     DIMENSION_ID_INCIDENT_DATE,
     DIMENSION_ID_SCHEDULED_DATE,
     DIMENSION_ID_LAST_UPDATED,
-    DIMENSION_ID_SCHEDULED_DATE,
 } from '../../src/modules/dimensionConstants.js'
 import { getPreviousYearStr } from '../helpers/period.js'
 
--- a/cypress/data/index.js
+++ b/cypress/data/index.js
@@ -1,10 +1,22 @@
 import {
+    DIMENSION_ID_EVENT_STATUS,
+    DIMENSION_ID_PROGRAM_STATUS,
+    DIMENSION_ID_CREATED_BY,
+    DIMENSION_ID_LAST_UPDATED_BY,
     DIMENSION_ID_EVENT_DATE,
     DIMENSION_ID_ENROLLMENT_DATE,
     DIMENSION_ID_INCIDENT_DATE,
+    DIMENSION_ID_SCHEDULED_DATE,
     DIMENSION_ID_LAST_UPDATED,
 } from '../../src/modules/dimensionConstants.js'
 import { getPreviousYearStr } from '../helpers/period.js'
+
+const mainDimensions = {
+    [DIMENSION_ID_EVENT_STATUS]: 'Event status',
+    [DIMENSION_ID_PROGRAM_STATUS]: 'Program status',
+    [DIMENSION_ID_CREATED_BY]: 'Created by',
+    [DIMENSION_ID_LAST_UPDATED_BY]: 'Last updated by',
+}
 
 export const ANALYTICS_PROGRAM = {
     programName: 'Analytics program',
@@ -12,7 +24,9 @@ export const ANALYTICS_PROGRAM = {
     [DIMENSION_ID_EVENT_DATE]: 'Event date (analytics)',
     [DIMENSION_ID_ENROLLMENT_DATE]: 'Enrollment date (analytics)',
     [DIMENSION_ID_INCIDENT_DATE]: 'Incident date (analytics)',
+    [DIMENSION_ID_SCHEDULED_DATE]: 'Scheduled date (analytics)',
     [DIMENSION_ID_LAST_UPDATED]: 'Last updated on',
+    ...mainDimensions,
 }
 
 export const HIV_PROGRAM = {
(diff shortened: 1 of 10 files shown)


martinkrulltott (commented):
LGTM, but needs Cypress tests for regression testing

cypress/integration/table.cy.js
```diff
@@ -0,0 +1,113 @@
...
+
+    expectTableToBeVisible()
+
+    cy.getBySelLike('layout-chip').contains(`${dimensionName}: all`)
+}
+
+describe('table', () => {
+    beforeEach(() => {
+        cy.visit('/', EXTENDED_TIMEOUT)
+        setUpTable()
```
martinkrulltott, line 72: If there's only 1 test (i.e. `it`) then you might as well skip the `beforeEach` and just do this at the start of the `it` instead.

cypress/integration/table.cy.js
```diff
@@ -0,0 +1,113 @@
...
+            .click({ force: true })
+
+        cy.contains('Add to Columns').click()
+    })
+
+    clickMenubarUpdateButton()
+
+    expectTableToBeVisible()
+
+    cy.getBySelLike('layout-chip').contains(`${dimensionName}: all`)
```
martinkrulltott, line 66: I don't think this is relevant to this test and should be removed. As per:

> Fast-forward approach: Skip ahead to the area you want to test and focus on adding assertions here, only test things along the way that are actually relevant to the test.

cypress/integration/table.cy.js
```diff
@@ -0,0 +1,113 @@
...
+        cy.visit('/', EXTENDED_TIMEOUT)
+        setUpTable()
+    })
+
+    it('click on column header opens the dimension dialog', () => {
+        // check the correct number of columns
+        getTableHeaderCells().its('length').should('equal', 11)
+
+        // check that there is at least 1 row in the table
+        getTableRows().its('length').should('be.gte', 1)
```
martinkrulltott, line 80: why is this relevant to the test?

cypress/integration/table.cy.js
```diff
@@ -0,0 +1,113 @@
...
+
+        // check the column headers in the table
+        getTableHeaderCells()
```
martinkrulltott, line 83: There's a lot of repetition here, 1. click the header, 2. check the modal title, 3. close the modal.
I would set this up as an array with e.g. ´{ label: X, expectedTitle: Y }` with all the cases / dimensions and then just loop through them all at once.

cypress/integration/table.cy.js
edoardo: 👍 removed.

cypress/integration/table.cy.js
edoardo: Removed.

cypress/data/index.js
```diff
@@ -1,18 +1,32 @@
...
+    DIMENSION_ID_LAST_UPDATED_BY,
     DIMENSION_ID_EVENT_DATE,
     DIMENSION_ID_ENROLLMENT_DATE,
     DIMENSION_ID_INCIDENT_DATE,
+    DIMENSION_ID_SCHEDULED_DATE,
     DIMENSION_ID_LAST_UPDATED,
 } from '../../src/modules/dimensionConstants.js'
 import { getPreviousYearStr } from '../helpers/period.js'
 
+const mainDimensions = {
```
martinkrulltott, line 14: I don't see the need to append the main dimension names to the different programs, as the names are static and won't change (compared to e.g. `eventDate` that is dynamic).
On further thought, I don't see the need to keep the `mainDimensions` here in `/data/index.js` either, as its only used in the one place (so far). No need to over-complicate things.

cypress/integration/table.cy.js
edoardo: The idea is to add more tests related to table in this file, we have at least sorting and pagination that can be added.

cypress/data/index.js
edoardo: Refactored.

cypress/integration/table.cy.js
edoardo: Refactored.

cypress/integration/table.cy.js
martinkrulltott: Yeah ok but will the other tests be using the same table setup? Seems a bit much to add all of the main dimensions just for sorting and pagination...

edoardo: Those are "problematic" columns because of the remapping going on, so I would assume we do want to test all of them for sorting for example.
Pagination does not have much to do with the table columns.

martinkrulltott: Ok fair enough. Of course feel free to refactor in the future if needed 👍

martinkrulltott (approved):

PR #215, opened 2022-09-20
fix: requesting data without a stage does not work (DHIS2-13764)

janhenrikoverland:
Fix for [DHIS2-13764](https://dhis2.atlassian.net/browse/DHIS2-13764)

- [x] Fix when update is clicked
- [x] Fix download
- [x] Add tests

After:
image


cypress/integration/createEventWithoutStage.cy.js
```diff
@@ -0,0 +1,27 @@
+import { DIMENSION_ID_EVENT_DATE } from '../../src/modules/dimensionConstants.js'
+import { AEFI_PROGRAM, TEST_REL_PE_LAST_12_MONTHS } from '../data/index.js'
+import { selectEventProgram } from '../helpers/dimensions.js'
+import { clickMenubarUpdateButton } from '../helpers/menubar.js'
+import { selectRelativePeriod } from '../helpers/period.js'
+import { expectTableToBeVisible } from '../helpers/table.js'
+import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+const program = AEFI_PROGRAM
```
martinkrulltott, line 9: Unless you think the `AEFI_PROGRAM` will be useful for other tests I wouldn't create a const for it, here's an example of a single-use of the `MAL-CS` program where I just wrote it in plain text to keep things simple https://github.com/dhis2/line-listing-app/blob/b75b27b911b93ad189d1a041c64e8822d3a860e5/cypress/integration/repeatedEvents.cy.js#L187

cypress/integration/createEventWithoutStage.cy.js
```diff
@@ -0,0 +1,27 @@
...
+const periodLabel = program[DIMENSION_ID_EVENT_DATE]
+
+describe('event without stage', () => {
```
martinkrulltott, line 12: This test is so short and simple I'd move it to the bottom of the `createEvent` file rather than keeping it in a file on its own...

janhenrikoverland: If we want multiple `describe`s per file then we can move it.

martinkrulltott: either multiple separate `describe` or we can wrap the two individual ones in an outer `describe` as well (see my suggestion here: https://github.com/dhis2/line-listing-app/pull/217)

martinkrulltott (approved):

PR #219, opened 2022-09-21
feat: add "Scheduled" event status (DHIS2-13881)

janhenrikoverland:
See commit log for a summary of feats and fixes in this release.


janhenrikoverland:
Use netlify build of last merged PR instead.

PR #238, opened 2022-10-05
test: assert value types in visualization

martinkrulltott:
Tests all value types in the visualization / table


janhenrikoverland (approved):
LG

PR #245, opened 2022-10-10
fix: reuse layout validation and sync filemenu/download

janhenrikoverland:
Adjust filemenu and download button to enable/disable correctly based on the state of current/visualization in the redux store.

### DIFF
--- a/src/modules/layoutValidation.js
+++ b/src/modules/layoutValidation.js
@@ -24,25 +24,27 @@ const isAxisValid = (axis) =>
         })
     )
 
-const validateDimension = (dimension, error, requireItems) => {
-    if (!(dimension && dimensionIsValid(dimension, { requireItems }))) {
-        throw error
+export const validateLineListLayout = (layout, { doNotThrow } = {}) => {
+    if (!layout) {
+        return false
     }
-}
 
-const validateAxis = (axis, error) => {
-    if (!isAxisValid(axis)) {
-        throw error
+    if (!isAxisValid(layout.columns)) {
+        if (doNotThrow) {
+            return false
+        }
+        throw noColumnsError()
     }
-}
 
-const validateLineListLayout = (layout) => {
-    validateAxis(layout.columns, noColumnsError())
-    validateDimension(
-        layoutGetDimension(layout, DIMENSION_ID_ORGUNIT),
-        noOrgUnitError(),
-        true
-    )
+    const ouDimension = layoutGetDimension(layout, DIMENSION_ID_ORGUNIT)
+    if (
+        !(ouDimension && dimensionIsValid(ouDimension, { requireItems: true }))
+    ) {
+        if (doNotThrow) {
+            return false
+        }
+        throw noOrgUnitError()
+    }
 
     let layoutHasTimeDimension = false
 
@@ -55,16 +57,27 @@ const validateLineListLayout = (layout) => {
     })
 
     if (!layoutHasTimeDimension) {
+        if (doNotThrow) {
+            return false
+        }
         throw noPeriodError()
     }
 
     if (!layoutHasProgramId(layout)) {
+        if (doNotThrow) {
+            return false
+        }
         throw noProgramError()
     }
 
     if (layout.outputType === OUTPUT_TYPE_EVENT && !layout?.programStage?.id) {
+        if (doNotThrow) {
+            return false
+        }
         throw noStageError()
     }
+
+    return true
 }
 
 export const validateLayout = (layout) => {
@@ -75,15 +88,9 @@ export const validateLayout = (layout) => {
     }
 }
 
-export const layoutHasProgramId = (layout) => {
-    if (!layout) {
-        return false
-    }
-    switch (layout.type) {
-        case VIS_TYPE_LINE_LIST:
-        default:
-            return Boolean(layout.program?.id)
-    }
-}
+export const layoutHasProgramId = (layout) => Boolean(layout?.program?.id)
 
 export const aoCreatedInEventReportsApp = (layout) => layout.legacy
+
+export const isLayoutValidForSaving = (layout) =>
+    layoutHasProgramId(layout) && !aoCreatedInEventReportsApp(layout)
--- a/src/components/DownloadMenu/useDownloadMenu.js
+++ b/src/components/DownloadMenu/useDownloadMenu.js
@@ -2,6 +2,7 @@ import { Analytics } from '@dhis2/analytics'
 import { useConfig, useDataEngine } from '@dhis2/app-runtime'
 import { useState, useCallback } from 'react'
 import { useSelector } from 'react-redux'
+import { validateLineListLayout } from '../../modules/layoutValidation.js'
 import { sGetCurrent } from '../../reducers/current.js'
 import {
     getAnalyticsEndpoint,
@@ -141,7 +142,7 @@ const useDownloadMenu = (relativePeriodDate) => {
     return {
         isOpen,
         toggleOpen: () => setIsOpen(!isOpen),
-        disabled: !current,
+        disabled: !validateLineListLayout(current, { doNotThrow: true }),
         download,
     }
 }
--- a/src/components/Toolbar/MenuBar/MenuBar.js
+++ b/src/components/Toolbar/MenuBar/MenuBar.js
@@ -13,10 +13,7 @@ import { tSetCurrent } from '../../../actions/current.js'
 import { acSetVisualization } from '../../../actions/visualization.js'
 import { getAlertTypeByStatusCode } from '../../../modules/error.js'
 import history from '../../../modules/history.js'
-import {
-    aoCreatedInEventReportsApp,
-    layoutHasProgramId,
-} from '../../../modules/layoutValidation.js'
+import { isLayoutValidForSaving } from '../../../modules/layoutValidation.js'
 import { getVisualizationFromCurrent } from '../../../modules/visualization.js'
 import { sGetCurrent } from '../../../reducers/current.js'
 import { sGetVisualization } from '../../../reducers/visualization.js'
@@ -207,6 +204,12 @@ const MenuBar = ({
         onError,
     })
 
+    const fileMenuObject = {
+        id: visualization?.id,
+        access: visualization?.access,
+        ...current,
+    }
+
     return (
         <div className={classes.menuBar} data-test="menubar">
             <UpdateVisualizationContainer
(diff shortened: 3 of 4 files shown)


src/modules/layoutValidation.js
```diff
@@ -24,25 +24,27 @@ const isAxisValid = (axis) =>
         })
     )
 
-const validateDimension = (dimension, error, requireItems) => {
-    if (!(dimension && dimensionIsValid(dimension, { requireItems }))) {
-        throw error
+export const validateLineListLayout = (layout, { doNotThrow } = {}) => {
```
martinkrulltott, line 27: instead of introducing a new prop (`doNotThrow`), couldn't you just catch the error that is being thrown and set a prop to to false instead?
for example something like this in `useDownloadMenu.js`
```
let something = true
try { 
    validateLineListLayout(current) 
} catch (error) {
    if (isVisualizationError(error)) {
        something = false
    } else {
        throw error
    }
}
```

janhenrikoverland: The idea is that the validation function now can be used for dry runs. We want to instruct the function when we call it.

martinkrulltott: oh ok, adding the dry-run option seems a bit over-engineered, when you can just catch the error in the consumer if you don't want it to throw. But if you insist then I would at least call it `dryRun` (or something similar) instead of `doNotThrow`...

janhenrikoverland: @martinkrulltott I missed the point of your first comment. I'll revisit it.

cypress/integration/download.cy.js
```diff
@@ -0,0 +1,38 @@
+import { DIMENSION_ID_EVENT_DATE } from '../../src/modules/dimensionConstants.js'
+import { HIV_PROGRAM, TEST_REL_PE_LAST_12_MONTHS } from '../data/index.js'
+import { selectEventProgram } from '../helpers/dimensions.js'
+import { clickMenubarUpdateButton } from '../helpers/menubar.js'
+import { selectRelativePeriod } from '../helpers/period.js'
+import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+describe('download', () => {
+    it('download button enable/disable', () => {
```
martinkrulltott, line 9:
```suggestion
    it('is only enabled when a program is selected', () => {
```

cypress/integration/translations.cy.js
```diff
@@ -1,7 +1,7 @@
 import { EXTENDED_TIMEOUT } from '../support/util.js'
 
-const UPDATE_BUTTON_ORIGINAL = 'Update'
-const UPDATE_BUTTON_TRANSLATED = 'Oppdater'
+const UPDATE_ITEM_ORIGINAL = 'Update'
```
martinkrulltott, line 3: Why "Item", it's referring to a button 🤷 
Just like "WELCOME_MSG" is referring to the message...

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,153 @@
...
+const ITEM_SAVE = 'file-menu-save'
+const ITEM_SAVEAS = 'file-menu-saveas'
+const ITEM_RENAME = 'file-menu-rename'
+const ITEM_TRANSLATE = 'file-menu-translate'
+const ITEM_SHARING = 'file-menu-sharing'
+const ITEM_GETLINK = 'file-menu-getlink'
+const ITEM_DELETE = 'file-menu-delete'
+
+const defaultItemsMap = {
+    [ITEM_NEW]: false,
```
martinkrulltott, line 17: Shouldn't `NEW` and `OPEN` always be `true`? They are indeed the "default" state as that's what you get without making any selection at all..

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,153 @@
...
+    [ITEM_SHARING]: false,
+    [ITEM_GETLINK]: false,
+    [ITEM_DELETE]: false,
+}
+
+const assertItems = (enabledItemsMap) => {
+    const itemsMap = Object.assign({}, defaultItemsMap, enabledItemsMap)
+
+    Object.keys(itemsMap).forEach((itemName) => {
+        itemsMap[itemName]
```
martinkrulltott, line 32: if you're accessing both the key (through the prop] and the value (through `itemsMap[x]`) you might as well use `Object.entries` that does this for you rather than `Object.keys` and looping through the array again...

martinkrulltott: i.e. 
```
Object.entries(itemsMap).forEach(([key, value]) => {
        value 
        ? cy.getBySel(key).should...
```

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,153 @@
...
+            ? cy.getBySel(itemName).should('not.have.class', 'disabled')
+            : cy.getBySel(itemName).should('have.class', 'disabled')
+    })
+}
+
+describe('file menu', () => {
+    it('reflects empty state', () => {
+        cy.visit('/', EXTENDED_TIMEOUT)
+
+        cy.getBySel('menubar').contains('File').click()
```
martinkrulltott, line 42: I would consider moving `cy.getBySel('menubar').contains('File').click()` into `assertItems` as they're always bundled together (you need to open the file menu before accessing it, so why not do both things at once?)

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,153 @@
...
+            [ITEM_RENAME]: true,
+            [ITEM_TRANSLATE]: true,
+            [ITEM_SHARING]: true,
+            [ITEM_GETLINK]: true,
+            [ITEM_DELETE]: true,
+        })
+
+        // cleanup
+        cy.getBySel(ITEM_DELETE).click()
+        cy.getBySel('file-menu-delete-modal-delete').click()
```
martinkrulltott, line 151: I would add some sort of assert after this step to make sure that it was indeed deleted.
Either looking for the "deleted" message and/or run the `assertItems` again to make sure the file menu was reset to "default".
Otherwise the delete here might fail silently without us ever knowing (i.e. rule of thumb, always end all tests with an assertion).

martinkrulltott (commented):
Tests look good in principle, some comments added for possible improvements though.

cypress/integration/fileMenu.cy.js
janhenrikoverland: I'm thinking that the delete function should have it's own test, instead of "hiding" it in a disable/enable test, and to avoid overlapping we should not test it again here. But I guess it's okay to test the frontend part of it, i.e. that the menu gets reset.

cypress/integration/fileMenu.cy.js
janhenrikoverland: Object.keys is faster, so keeping it.

cypress/integration/translations.cy.js
janhenrikoverland: Unintentional, replace all mistake probably.

src/modules/layoutValidation.js
janhenrikoverland: Ok, I'll rename to `dryRun`. Technically the function doesn't return errors, so the dry run returns true or false while a normal run returns true or throws an error. I think it's nice to be able to do a dry run without trycatching.

cypress/integration/fileMenu.cy.js
janhenrikoverland: Updated

cypress/integration/fileMenu.cy.js
janhenrikoverland: Updated

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,140 @@
...
+        cy.getBySel('file-menu-saveas-modal-save').click()
+
+        cy.getBySel('visualization-title').contains(name)
+
+        clickMenubarUpdateButton()
+
+        cy.getBySel('visualization-title').contains('Edited')
+
+        assertItems({
+            [ITEM_NEW]: true,
```
martinkrulltott, line 121: `ITEM_NEW` and `ITEM_OPEN` are redundant here, can be removed (already `true` by default)

cypress/integration/fileMenu.cy.js
```diff
@@ -0,0 +1,140 @@
...
+            [ITEM_SAVE]: true,
+        })
+    })
+
+    it('reflects saved state', () => {
+        cy.visit('/#/ZTrsv19jw9U', EXTENDED_TIMEOUT)
+
+        cy.getBySel('visualization-title').contains('COVAC enrollment')
+
+        assertItems({
```
martinkrulltott, line 66: shouldn't `SAVE` be `true` here as well? If it's opening an AO where saving is allowed? (compared to the test below, where saving isn't allowed)

janhenrikoverland: No, SAVE is only allowed when it's `dirty`. Otherwise use SAVE AS.

cypress/integration/fileMenu.cy.js
janhenrikoverland: Updated

cypress/integration/fileMenu.cy.js
janhenrikoverland: Updated

cypress/integration/download.cy.js
```diff
@@ -0,0 +1,74 @@
...
+        clickMenubarUpdateButton()
+
+        downloadIsDisabled()
+
+        selectRelativePeriod({
+            label: HIV_PROGRAM[DIMENSION_ID_EVENT_DATE],
+            period: TEST_REL_PE_LAST_12_MONTHS,
+        })
+
+        clickMenubarUpdateButton()
```
martinkrulltott, line 47: I would add a `expectTableToBeVisible()` between the update and the download being enabled, as I assume that's what ultimately determines if you can download the table, right? (e.g. if for some reason the steps taken above isn't enough to generate a table, then the button should be disabled still and the test should fail)

cypress/integration/fileMenu.cy.js
janhenrikoverland: Fixed now

cypress/integration/download.cy.js
janhenrikoverland: I think it's okay to enable the download button as soon as the required steps for passing the layout validation are met. In the app, whether analytics returned data or not e.g. for the selected period is kind of out of our hands. We could potentially look at `store.loader.loadingError` in combination with `store.current`, but not sure if we want to go down that road.

martinkrulltott (approved):

PR #248, opened 2022-10-12
chore: run e2e tests against all supported versions of the api

jenniferarnesen:
The `e2e-tests` workflow will run the cypress tests against supported DHIS2 stable production instances and dev.
The `verify-app` workflow no longer runs e2e tests, focusing instead on lint, jest tests and building the release artifact.

- the `e2e-tests` workflow
    - runs: 
        - on push to our main branch (currently `dev` but in the future it will be `master`)
        - on pull requests. In addition it only runs if the PR is not a draft (and it is not from a fork)
    - determines all the "legacy" versions (supported dhis2 versions in prod) and then runs the tests against those versions.  Legacy naming convention is `2.38lytics`, `2.39lytics`... It only tests the latest stable version, not each patch. So if a 2.38.0, 2.38.1 and 2.38.2 all exist, it will only test against 2.38.2.
    - runs the tests against "dev" (separate job in the workflow)
    - declares success if the legacy versions pass. If the dev test run fails, this does not prevent merging the PR
    - the tests rely on the `dhis2InstanceVersion` being set.  On ci this is handled automatically. But when running the tests locally you need to set it in cypress.env.json (explained in the README)

- tags
    - we can tag tests to only run on certain versions. This is explained in the README
    - plugin `tagify` from cypress-tags is used for the tagging (`typescript` had to be added as a dependency)
   
- yarn scripts
    - removed unused `cypress:capture` and `cypress:stub`. Added `cy:run` (for running non-interactive locally), and renamed `cypress:open` to `cy:open`

- tags have been added to legendSet and event cypress tests so that they don't run on 2.38

- the `getExcludedTags` function calculates which tags should be excluded from the test run based on the instance version and minimum version. Meaning, a test with an excluded tag will NOT be run. Jest tests for this function have been added to hopefully clarify the expected output of this function

TODO: after this PR is merged, then add the `e2e-tests-success` job as required for merging (in the repo settings)


.github/workflows/dhis2-verify-app.yml
```diff
@@ -75,11 +63,6 @@ jobs:
             - uses: c-hive/gha-yarn-cache@v1
             - run: yarn install --frozen-lockfile
 
-            # Can be removed if translations aren't required for tests,
-            # or if not using the App Platform
-            - name: Generate translations
-              run: yarn d2-app-scripts i18n generate
```
martinkrulltott, line 81: is this safe to remove, given that we're using translations in our tests? https://github.com/dhis2/line-listing-app/blob/4bc7bad81fbdd4ef0b3c2a7a0ab03228ff2d84fd/cypress/integration/translations.cy.js#L37

jenniferarnesen: Yes bc this is only for Jest tests.

The cypress-io/github-action does all the installing and building needed to run the e2e tests (as you can see they are passing)

package.json
```diff
@@ -25,14 +24,16 @@
...
+        "cypress-tags": "^1.1.2",
         "enzyme": "^3.11.0",
         "enzyme-adapter-react-16": "^1.15.6",
         "eslint-plugin-cypress": "^2.12.1",
         "identity-obj-proxy": "^3.0.0",
         "jest-enzyme": "^7.1.2",
         "redux-mock-store": "^1.5.4",
-        "start-server-and-test": "^1.14.0"
+        "start-server-and-test": "^1.14.0",
+        "typescript": "^4.8.4"
```
martinkrulltott, line 36: why was this added?

jenniferarnesen: Because cypress-tags requires that

cypress/support/getExcludedTags.js
```diff
@@ -0,0 +1,74 @@
+const getString = (v) => (typeof v === 'number' ? v.toString() : v)
+
+const extractMinorVersion = (v) =>
+    v.indexOf('2.') === 0 ? parseInt(v.slice(2, 4)) : parseInt(v.slice(0, 2))
```
martinkrulltott, line 4: this looks like its made to support both `2.40` and just `40`. All examples of tags use the `xx` format, but dhis2 instances use the `2.xx` format. Should we maybe decide on which format to use to keep things consistent? or should the support for both count as a feature? 🙂

jenniferarnesen: I included both to offer some flexibility, but maybe it's better to just be strict about it from the get-go. I'm pretty sure the instance version will always be `2.xx` on CI. So we could probably simplify and enforce `2.xx`.

martinkrulltott (approved):

cypress/integration/eventStatus.cy.js
```diff
@@ -103,4 +110,38 @@ describe('event status', () => {
 
         assertTooltipContainsEntries(['Scheduled'])
     })
+
+    it('can be filtered by status ACTIVE', () => {
+        setUpTable(event[DIMENSION_ID_LAST_UPDATED])
```
martinkrulltott, line 115: This in combination with 
``` expectTableToMatchRows([`${getCurrentYearStr()}-05-03`]) ```
below is causing the test to fail when the row has been updated in the db, as the date is updated dynamically. I'll switch this to use one of the static periods instead (event date etc)

PR #258, opened 2022-10-18
fix: scheduled date being hidden by program stage not respected (DHIS2-13904)

martinkrulltott:
Implements [DHIS2-13904](https://dhis2.atlassian.net/browse/DHIS2-13904)

---

### Key features

1. Hide schedule date based on stage setting

---

### Screenshots

image

### DIFF
--- a/cypress/integration/timeDimensions.cy.js
+++ b/cypress/integration/timeDimensions.cy.js
@@ -6,7 +6,11 @@ import {
     DIMENSION_ID_SCHEDULED_DATE,
 } from '../../src/modules/dimensionConstants.js'
 import { ANALYTICS_PROGRAM, TEST_REL_PE_THIS_YEAR } from '../data/index.js'
-import { selectEventProgram } from '../helpers/dimensions.js'
+import {
+    dimensionIsDisabled,
+    dimensionIsEnabled,
+    selectEventProgram,
+} from '../helpers/dimensions.js'
 import { clickMenubarUpdateButton } from '../helpers/menubar.js'
 import { selectRelativePeriod } from '../helpers/period.js'
 import {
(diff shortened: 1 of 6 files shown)


src/actions/ui.js
```diff
@@ -122,7 +123,11 @@ export const tClearUiProgramStageDimensions =
         if (needsClearing) {
             const idsToRemove = state.ui.layout.columns
                 .concat(state.ui.layout.filters)
-                .filter((id) => id.includes(`${stageId}.`))
+                .filter(
+                    (dimensionId) =>
+                        dimensionId.includes(`${stageId}.`) ||
+                        dimensionId === DIMENSION_ID_SCHEDULED_DATE
```
janhenrikoverland, line 129: This is fine for now, but we should ideally centralize all the "rules" somewhere to make them easy to keep track of and come back and update later.

janhenrikoverland (approved):

cypress/integration/timeDimensions.cy.js
```diff
@@ -56,4 +60,50 @@ describe('event', () => {
                 .should('be.visible')
         })
     })
+    it('scheduled date disabled state is set based on stage setting ', () => {
```
jenniferarnesen, line 63: Should this test be tagged '>=39' ?

martinkrulltott: Yes. And with that said, there will be a lot of tests failing in 2.38 as it seems right now, everything using `DIMENSION_ID_SCHEDULED_DATE` will fail since it's not available...

jenniferarnesen: Will be easier to spot these as soon as we get a real 2.38...

PR #260, opened 2022-10-21
test: feature-toggle tests

martinkrulltott:
Adds tags to toggle tests based on the feature toggling done by the app.
2.39 adds support for `scheduleDate`, which doesn't exist in 2.38, so all tests related to `scheduleDate` are toggled with the `>=39` tag. 
Since these tests loop through multiple items, the tags are added at a high level and subsequently pass down props to the actual tests for them to toggle the relevant part, to keep things as DRY as possible.


cypress/integration/table.cy.js
```diff
@@ -153,77 +206,43 @@ const assertDimensions = (inputDimensions) => {
...
+    // remove org unit
+    cy.getBySel('layout-chip-ou').findBySel('dimension-menu-button').click()
+    cy.containsExact('Remove').click()
+}
 
-            cy.contains('Add to Columns').click()
+describe(['<40'], 'table', () => {
+    beforeEach(init)
+    it('click on column header opens the dimension dialog', () => {
+        programDimensions.push({
```
jenniferarnesen, line 220: I would say that we just don't test this since the value is a bug.

cypress/integration/table.cy.js
```diff
@@ -153,77 +206,43 @@ const assertDimensions = (inputDimensions) => {
...
+})
 
-        selectFixedPeriod({
-            label: periodLabel,
-            period: {
-                type: 'Daily',
-                year: `${getPreviousYearStr()}`,
-                name: `${getPreviousYearStr()}-12-10`,
-            },
+describe(['dev', '>=40'], 'table', () => {
```
jenniferarnesen, line 231: The 'dev' tag can be removed for now. If we change the way we do tags, we can always add it in later. Also, could you move the comment on line 232 down to line 239 since it is only relevant to the program dimension, not scheduled date?

jenniferarnesen (approved):

cypress/integration/table.cy.js
jenniferarnesen: Also, the test is failing:  
image

cypress/integration/createEnrollment.cy.js
```diff
@@ -28,7 +28,8 @@ const enrollment = ANALYTICS_PROGRAM
 const dimensionName = TEST_DIM_TEXT
 const periodLabel = enrollment[DIMENSION_ID_ENROLLMENT_DATE]
 
-const setUpTable = () => {
+const setUpTable = ({ scheduleDateIsSupported } = {}) => {
+    console.log(`scheduleDateIsSupported: ${scheduleDateIsSupported}`)
```
jenniferarnesen, line 32: can the console.log be removed?

PR #263, opened 2022-10-26
ci: use reusable verify-commits workflow

radnov:


radnov:
Closing this PR, as a more thought-through approach would be needed to reuse GHA workflows across the app repos.

PR #270, opened 2022-11-04
test: add intercept and customize interpretation text with current time in cypress test

jenniferarnesen:
The edit interpretation test sometimes fails, and it's unclear why. Added POST and PUT intercept so that it is confirmed that the request was successful.


edoardo:
The interpretations tests have been rewritten to create a new AO every time they run, and delete it afterwards.
This should avoid the failures in case something goes wrong in a run and the AO remains in a dirty state.

jenniferarnesen:
Not needed after interpretation tests were refactored

PR #275, opened 2022-11-08
chore: enable program dimensions tests by expanding structure of test program data

jenniferarnesen:
The impetus for this refactor was to enable all the tests in programDimensions.cy.js. (there was a `.only` on one of the tests in that file). In order to get all the tests to work, the enabled/disabled state of the time dimensions for the given program/stage was needed. This triggered a restructure of ANALYTICS_PROGRAM and HIV_PROGRAM. This change caused cascading changes in other files that use those objects.

In the future we could put the other programs in use in our tests into this file.

An additional refactor could be to create functions for getting the time dimension labels and enabled state so that if the object structure changes again, it won't cause a cascading effect like this.
e.g `event.stages[stageName][DIMENSION_ID_EVENT_DATE].label` becomes `getExpectedTimeDimensionLabel`

### DIFF
--- a/cypress/data/index.js
+++ b/cypress/data/index.js
@@ -60,31 +60,6 @@ export const ENROLLMENT_ANALYTICS_PROGRAM = {
     },
 }
 
-export const ENROLLMENT_HIV_PROGRAM = {
-    programName: 'HIV Case Surveillance',
-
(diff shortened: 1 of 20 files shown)


jenniferarnesen:
Decided not to take this approach.

PR #291, opened 2022-11-25
test: align cypress tests with new db

janhenrikoverland:
WIP


edoardo (approved):

cypress/integration/conditions/booleanConditions.cy.js
```diff
@@ -73,11 +73,12 @@ describe('boolean conditions - Yes/NA', () => {
         addConditions(['Not answered'], dimensionName)
 
         expectTableToMatchRows([
+            `${currentYear}-04-19`,
```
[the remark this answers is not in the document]
janhenrikoverland: Yes

cypress/integration/eventStatus.cy.js
```diff
@@ -49,13 +49,12 @@ describe('event status', () => {
         setUpTable(event[DIMENSION_ID_SCHEDULED_DATE])
 
         expectTableToMatchRows([
-            'Active',
-            'Scheduled',
```
martinkrulltott, line 53: If there's no `scheduled` item in the db any longer then what's the point of this test?

cypress/integration/repeatedEvents.cy.js
```diff
@@ -81,124 +81,96 @@ describe('repeated events', () => {
...
-            '',
-            '5',
-            '',
-        ]
+        const dimensionName = 'E2E - Percentage'
+        setUpTable({ enrollment: E2E_PROGRAM, dimensionName })
+
+        // repetition 6/0 can be set successfully
+        setRepetition({ dimensionName, recent: 6, oldest: 0 })
+        const result = ['', '', '', '', '45', '46']
```
martinkrulltott, line 153: previously there was 9 rows and the `10/0` was there to test that the result contained all 9 plus an empty out of bounds result as well. This can't be tested with only 2 rows in the db, so this test now doesn't do its duty..

martinkrulltott (commented):
Reviewed, see separate comments above

PR #297, opened 2022-12-06
fix: clear repetition when changing input type (DHIS2-14219)

janhenrikoverland:
Implements [DHIS2-14219](https://dhis2.atlassian.net/browse/DHIS2-14219)

Clears the ui.repetition config when switching input. I did not want to clear it as part of clearing the program in case that becomes an issue when we support cross-program selection.


janhenrikoverland:
> Unrelated to the specific fix, but I found a bit confusing `acSetUiRepetition` after seeing we have both `acRemoveUiRepetition` and `acClearUiRepetition`. The set seems to in fact add a repetition, so perhaps is more clear if it's renamed to `acAddUiRepetition` which works with `acRemoveUiRepetition`...

Agreed. We've had a pattern where "set" means replacing the entire object. While for appending to a list we've used "add". So that makes sense.

PR #300, opened 2022-12-09
fix: use more specific selectors in cypress test to avoid flakiness

jenniferarnesen:
dateConditions:
- there was a vis that had "Last 12 months" in the name, that was listed in the start screen behind the period selector dialog. Cypress was choosing that rather than the period option

fileMenu:
- on CI, the File menu was positioned at 0,0 in the viewport, so clicking on 0,0 didn't close the file menu. Instead, click on the cover layer to close the menu, and confirm that it's gone. I'm not sure why the File menu was positioned incorrectly.

### DIFF
--- a/cypress/helpers/period.js
+++ b/cypress/helpers/period.js
@@ -26,7 +26,10 @@ const selectRelativePeriod = ({ label, period }) => {
             .containsExact(period.type)
             .click()
     }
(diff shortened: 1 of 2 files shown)


janhenrikoverland:
Added to https://github.com/dhis2/line-listing-app/pull/298

PR #302, opened 2022-12-09
chore: always use headless Chrome for proper video support

edoardo:
### Key features

1. solve broken video recording on dev tests

---

### Description

Electron browser seem to have issues with recording videos of test failures.
Chrome on the other hand works fine.
Always use Chrome for running tests in both production and dev test groups.


janhenrikoverland (approved):
LG

PR #305, opened 2022-12-12
fix: cypress - wait for options request to complete

jenniferarnesen:
This test keeps breaking.

Ensure that the options request is complete before trying to select from the source list


edoardo:
🤞

jenniferarnesen:
Solving issue by checking that the result of the search is shown in ui, instead of intercept

PR #307, opened 2022-12-13
fix: filtering by stage resets the search and type filters (DHIS2-14288)

martinkrulltott:
Implements [DHIS2-14288](https://dhis2.atlassian.net/browse/DHIS2-14288)

---

### Key features

1. split stage selection by input type to `StageSelect` and `StageFilter`

---

### Description

There are two distinct cases for selecting a stage

1. For _input type: Event_ stage is selected together with selecting a program, which affects the whole AO
2. For _input type: Enrollment_ stage can be used as a filter when browsing data elements, which only affects the list of dimensions (comparable to searching using the text field)

The `StageSelect` component was previously used for both of these use cases, which created a bit of a mess both in the code and in the ux. There has previously been a need to stop the "false" stage (_case 2_ above) from being passed to the backend etc, which patched up some of the issues with consolidating two distinct use cases into one component. But as there's now an issue with the filter itself (see JIRA ticket for more info), it was decided to split this one component into two - `StageSelect` for _case 1_ and `StageFilter` for _case 2_.

`StageSelect` still relies on the stage id in the store (`ui.program.stage`), but `StageFilter` is now using its own local state prop (called `stageFilter` / `setStageFilter`) to create a distinct separation.

---

### TODO

- [ ] Cypress tests

---

### Screenshots

_case 1_
image

_case 2_
image

### DIFF
--- a/src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsFilter.js
+++ b/src/components/MainSidebar/ProgramDimensionsPanel/ProgramDimensionsFilter.js
@@ -17,7 +17,7 @@ import { OUTPUT_TYPE_ENROLLMENT } from '../../../modules/visualization.js'
 import { sGetUiInputType } from '../../../reducers/ui.js'
 import { DimensionIcon } from '../DimensionItem/DimensionIcon.js'
 import styles from './ProgramDimensionsFilter.module.css'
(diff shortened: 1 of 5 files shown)


janhenrikoverland (approved):
Approved, add cypress tests after team testing.

PR #308, opened 2022-12-13
fix: digit group separator applied to valueType TEXT data elements (DHIS2-14172)

martinkrulltott:
Implements [DHIS2-14172](https://dhis2.atlassian.net/browse/DHIS2-14172)

---

### Key features

1. Stop applying DGS to optionSet options

---

### Description

Values that are actually options from an optionSet shouldn't have digit group separators applied to them, as those values doesn't count as being numeric. This fix prevents any value with an `optionSet` prop from getting formatted with DGS.

---

### TODO

-   [ ] Cypress tests

---

### Screenshots

_before_
image

_after_
image

### DIFF
--- a/cypress/integration/conditions/optionSetCondition.cy.js
+++ b/cypress/integration/conditions/optionSetCondition.cy.js
@@ -85,8 +85,8 @@ const assertNumericOptionSet = ({
 describe('Option set condition', () => {
     it(['>37', '<39'], 'Option set (number) displays correctly (2.38)', () => {
         assertNumericOptionSet({
-            tableFilteredOptionName: 'Ei ght',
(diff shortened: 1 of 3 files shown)


edoardo (commented):
Do we always want to use `formatValue`?
If we do one of these code blocks is executed:
https://github.com/dhis2/analytics/blob/master/src/modules/renderValue.js#L58-L60
https://github.com/dhis2/analytics/blob/master/src/modules/renderValue.js#L76-L83

Both are changing the value.
The former shouldn't be a problem, but the latter is more invasive.

jenniferarnesen:
Since our jest test file is pretty thorough, could you add a test for the new scenario, to keep it thorough?

edoardo (approved):

PR #314, opened 2022-12-21
fix: reset page on update (DHIS2-13950)

janhenrikoverland:
Implements [DHIS2-13950](https://dhis2.atlassian.net/browse/DHIS2-13950)

Resets the page to 1 for new visualizations. Stores the previous visualization in a ref and compares against the prop.

Example: we're on page 5 of a line list. If we listen to `visualization` and do `setPage` when changes, then the data hook first requests data for the new visualization with page 5, which we don't need and want to avoid. When local state for `page` is updated the data hook fires again, this time with page 1.

With this PR the first request is correct and the data engine cancels the second request because it's identical to the first one.

[DHIS2-13950]: https://dhis2.atlassian.net/browse/DHIS2-13950?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Visualization/Visualization.js
+++ b/src/components/Visualization/Visualization.js
@@ -18,7 +18,7 @@ import {
 } from '@dhis2/ui'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useRef, useCallback } from 'react'
 import {
     DISPLAY_DENSITY_COMFORTABLE,
     DISPLAY_DENSITY_COMPACT,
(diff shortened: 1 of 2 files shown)


src/components/Visualization/Visualization.js
```diff
@@ -79,19 +79,37 @@ export const Visualization = ({
         }
     )
 
-    const isInModal = !!filters?.relativePeriodDate
+    const visualizationRef = useRef(visualization)
+
+    const setPage = useCallback(
```
edoardo, line 84: Is this `useCallback` needed?
There isn't any logic using those dependencies, they are just passed to the `setSorting`, so in practice I don't think there is any advantage in using `useCallback`.

edoardo: Or is that needed because `setPage` is a dependency in a `useEffect`? 🤔

edoardo (approved):

src/components/Visualization/Visualization.js
janhenrikoverland: @edoardo Yes, otherwise there's an endless loop.

PR #317, opened 2022-12-29
fix: numbers with dgs space wrap on multiple lines (DHIS2-13899)

janhenrikoverland:
Implements [DHIS2-13899](https://dhis2.atlassian.net/browse/DHIS2-13899)

- Adds `white-space: nowrap` to cells according to ticket description.
- Adds a few test dims.
- Renames a few test dims to match value type name.
- Makes a very small change to display density (right padding) due to the nowrap css, very rarly visible to the user (ref https://dhis2.atlassian.net/browse/DHIS2-14303)

- [x] Add cypress test

Before:
Screenshot from 2022-12-29 11-37-26

After:
Screenshot from 2022-12-29 11-37-22

### DIFF
--- a/cypress/integration/value.cy.js
+++ b/cypress/integration/value.cy.js
@@ -62,7 +62,6 @@ describe('value', () => {
                 .invoke('css', 'white-space')
                 .should('equal', value)
 
-        console.log(getTableDataCells().all())
         ;[0, 1, 2, 3].forEach((index) => shouldHaveWhiteSpace(index, 'normal'))
         ;[4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15].forEach((index) =>
             shouldHaveWhiteSpace(index, 'nowrap')
--- a/cypress/integration/value.cy.js
+++ b/cypress/integration/value.cy.js
@@ -20,31 +20,39 @@ import {
 import { selectEventWithProgramDimensions } from '../helpers/dimensions.js'
 import { clickMenubarUpdateButton } from '../helpers/menubar.js'
 import { selectRelativePeriod } from '../helpers/period.js'
+import { goToStartPage } from '../helpers/startScreen.js'
 import { getTableDataCells } from '../helpers/table.js'
-import { EXTENDED_TIMEOUT } from '../support/util.js'
 
 describe('value', () => {
-    beforeEach(() => {
-        cy.visit('/', EXTENDED_TIMEOUT)
+    it('has the correct white-space css', () => {
+        goToStartPage()
+
+        const programDimensionsWithWrap = [
+            TEST_DIM_TEXT,
+            TEST_DIM_TEXT_OPTIONSET,
+            TEST_DIM_NUMBER_OPTIONSET,
+        ]
+        const programDimensionsWithoutWrap = [
+            TEST_DIM_NUMBER,
+            TEST_DIM_INTEGER,
+            TEST_DIM_INTEGER_POSITIVE,
+            TEST_DIM_INTEGER_NEGATIVE,
+            TEST_DIM_INTEGER_ZERO_OR_POSITIVE,
+            TEST_DIM_PERCENTAGE,
+            TEST_DIM_UNIT_INTERVAL,
+            TEST_DIM_TIME,
+            TEST_DIM_DATE,
+            TEST_DIM_DATETIME,
+            TEST_DIM_PHONE_NUMBER,
+        ]
+        const programDimensions = [
+            ...programDimensionsWithWrap,
+            ...programDimensionsWithoutWrap,
+        ]
 
         selectEventWithProgramDimensions({
             programName: E2E_PROGRAM.programName,
-            dimensions: [
-                TEST_DIM_TEXT,
-                TEST_DIM_TEXT_OPTIONSET,
-                TEST_DIM_NUMBER_OPTIONSET,
-                TEST_DIM_NUMBER,
-                TEST_DIM_INTEGER,
-                TEST_DIM_INTEGER_POSITIVE,
-                TEST_DIM_INTEGER_NEGATIVE,
-                TEST_DIM_INTEGER_ZERO_OR_POSITIVE,
-                TEST_DIM_PERCENTAGE,
-                TEST_DIM_UNIT_INTERVAL,
-                TEST_DIM_TIME,
-                TEST_DIM_DATE,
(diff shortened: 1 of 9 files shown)


cypress/integration/value.cy.js
```diff
@@ -0,0 +1,71 @@
...
+} from '../data/index.js'
+import { selectEventWithProgramDimensions } from '../helpers/dimensions.js'
+import { clickMenubarUpdateButton } from '../helpers/menubar.js'
+import { selectRelativePeriod } from '../helpers/period.js'
+import { getTableDataCells } from '../helpers/table.js'
+import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+describe('value', () => {
+    beforeEach(() => {
+        cy.visit('/', EXTENDED_TIMEOUT)
```
martinkrulltott, line 28:
```suggestion
        goToStartPage()
```

cypress/integration/value.cy.js
```diff
@@ -0,0 +1,71 @@
...
+    })
+
+    it('has the correct white-space css', () => {
+        const shouldHaveWhiteSpace = (index, value) =>
+            getTableDataCells()
+                .eq(index)
+                .invoke('css', 'white-space')
+                .should('equal', value)
+
+        console.log(getTableDataCells().all())
```
martinkrulltott, line 65: why `console.log`?

cypress/integration/value.cy.js
```diff
@@ -0,0 +1,71 @@
...
+    TEST_REL_PE_LAST_12_MONTHS,
+} from '../data/index.js'
+import { selectEventWithProgramDimensions } from '../helpers/dimensions.js'
+import { clickMenubarUpdateButton } from '../helpers/menubar.js'
+import { selectRelativePeriod } from '../helpers/period.js'
+import { getTableDataCells } from '../helpers/table.js'
+import { EXTENDED_TIMEOUT } from '../support/util.js'
+
+describe('value', () => {
+    beforeEach(() => {
```
martinkrulltott, line 27: why `beforeEach` when there's only a single `it`?

cypress/integration/value.cy.js
```diff
@@ -0,0 +1,71 @@
...
+
+    it('has the correct white-space css', () => {
+        const shouldHaveWhiteSpace = (index, value) =>
+            getTableDataCells()
+                .eq(index)
+                .invoke('css', 'white-space')
+                .should('equal', value)
+
+        console.log(getTableDataCells().all())
+        ;[0, 1, 2, 3].forEach((index) => shouldHaveWhiteSpace(index, 'normal'))
```
martinkrulltott, line 65: It's really difficult to understand why `0, 1, 2, 3` should be treated differently from the rest, especially since the actual table is generated in the `beforeEach` so it's quite disconnected from what's going on here.
Also tying the exact index of items in an array to the test result makes it quite fragile (adding something to the list above will mess up the test).

I would suggest to remove the `beforeEach` and keep everything within the `it`. Then, instead of relying on the exact index value, create a list of all dimensions that should have `nowrap` and all with `wrap`, then use `indexOf` to find them in the list.. 

e.g.

```

const dimensionsWithWrap = [TEST_DIM_TEXT, TEST_DIM_TEXT_OPTIONSET, TEST_DIM_NUMBER_OPTIONSET]
const dimensionsWithoutWrap = [......]
const dimensions =  [...dimensionsWithWrap, ...dimensionsWithoutWrap]

selectEventWithProgramDimensions({
            programName: E2E_PROGRAM.programName,
            dimensions,
...


dimensionsWithWrap.forEach((dimension) => shouldHaveWhiteSpace(dimensions.indexOf(dimension), 'normal'))
dimensionsWithoutWrap.forEach((dimension) => shouldHaveWhiteSpace(dimensions.indexOf(dimension), 'nowrap'))

```

This way you're not relying on exact indexes (less fragile) and the testes are much easier to read and understand what's going on.

src/components/Visualization/styles/Visualization.module.css
```diff
@@ -83,10 +83,10 @@
     padding: 8px 4px 6px 8px;
 }
 .dataTable .headerCell.sizeNormal {
-    padding: 4px 2px 2px 6px;
+    padding: 4px 4px 2px 6px;
```
martinkrulltott, line 86: what are all of these changes and why aren't they mentioned in the PR description?
Screenshots detailing this would be great as well. A ticket number would help as well.

cypress/integration/value.cy.js
martinkrulltott: This was marked as resolved but I don't see any code changes or comments.. (maybe Github is acting up 🤷 )

cypress/integration/value.cy.js
janhenrikoverland: I wanted to include the orgunit and event date, which are not passed as dimensions to the select.. function. But the orgunit is just TEXT and event date is DATE which we already test, so we can skip those which makes it less complex.

src/components/Visualization/styles/Visualization.module.css
janhenrikoverland: Description updated.

cypress/integration/value.cy.js
```diff
@@ -0,0 +1,84 @@
...
+} from '../data/index.js'
+import { selectEventWithProgramDimensions } from '../helpers/dimensions.js'
+import { clickMenubarUpdateButton } from '../helpers/menubar.js'
+import { selectRelativePeriod } from '../helpers/period.js'
+import { goToStartPage } from '../helpers/startScreen.js'
+import { getTableDataCells } from '../helpers/table.js'
+
+describe('value', () => {
+    it('has the correct white-space css', () => {
+        goToStartPage()
```
martinkrulltott, line 28: can be moved to below `programDimensions`

cypress/integration/value.cy.js
```diff
@@ -0,0 +1,84 @@
...
+        })
+
+        selectRelativePeriod({
+            label: E2E_PROGRAM[DIMENSION_ID_EVENT_DATE],
+            period: TEST_REL_PE_LAST_12_MONTHS,
+        })
+
+        clickMenubarUpdateButton()
+
+        const shouldHaveWhiteSpace = (index, value) =>
```
martinkrulltott, line 65: can be moved outside of the `it` or even the `describe`

martinkrulltott (approved):

PR #322, opened 2023-01-17
fix: do not pre-set program stage for enrollment DHIS2-14396

edoardo:
Implements [DHIS2-14396](https://dhis2.atlassian.net/browse/DHIS2-14396)

---

### Key features

1. do not set program stage id for enrollments
2. remove the stage from the download request for enrollments

---

### Description

Programs with only one stage get the stage automatically selected.
This only make sense when input type is event.
For enrollments we don't pass the stage to the analytics request, this fix avoids that the stage id is stored in the Redux store.

Cypress tests have been adjusted to not look for the program in the disabled time dimension labels.
These dimensions cannot be used with Enrollment input type, so the labels does not make any difference.
It was actually wrong that those dimensions got the program in the label, as for Enrollment multiple programs can be used.

---

### Screenshots

Before, program stage id was set in Redux even for enrollments:
Screenshot 2023-01-17 at 14 12 48

After, only the program id is set:
Screenshot 2023-01-17 at 14 13 14

[DHIS2-14396]: https://dhis2.atlassian.net/browse/DHIS2-14396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/DownloadMenu/useDownloadMenu.js
+++ b/src/components/DownloadMenu/useDownloadMenu.js
@@ -3,6 +3,7 @@ import { useConfig, useDataEngine } from '@dhis2/app-runtime'
 import { useState, useCallback } from 'react'
 import { useSelector } from 'react-redux'
 import { validateLineListLayout } from '../../modules/layoutValidation.js'
+import { OUTPUT_TYPE_ENROLLMENT } from '../../modules/visualization.js'
 import { sGetCurrent } from '../../reducers/current.js'
 import {
     getAnalyticsEndpoint,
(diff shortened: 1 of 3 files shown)


src/components/DownloadMenu/useDownloadMenu.js
```diff
@@ -121,6 +120,10 @@ const useDownloadMenu = (relativePeriodDate) => {
             // TODO add common parameters
             // if there are for both event/enrollment and PT/LL
 
+            if (current.outputType !== OUTPUT_TYPE_ENROLLMENT) {
```
martinkrulltott, line 123: might be more transparent to look for `=== OUTPUT_TYPE_EVENT` rather than the opposite?

martinkrulltott (approved):

src/components/DownloadMenu/useDownloadMenu.js
edoardo: The same condition is used in other places.
I believe the reason is because there will be more input types in the future, but I agree it seems a bit premature to think all other types can be treated as events in this specific case.

PR #339, opened 2023-03-28
test: add EXTENDED_TIMEOUT to all references of program-dimensions-list

martinkrulltott:
Recently the nightly test run has failed due to the program dimensions list being slow and not rendering within the default 4s timeout. This extends the timeout in all places where program-dimensions-list is used to avoid flakey tests.


jenniferarnesen (approved):

martinkrulltott:
The failing test for `repeatedEvents.cy.js` is caused by a missing AO, which should be present but seems to have been removed from the database

PR #345, opened 2023-04-20
chore: cypress upgrade using `cy.session` (do not merge)

HendrikThePendric:
I'm closing this PR in favor of #374. See that PR's description for a full explanation. 

I have included a summary of the last 21 runs of this PR below:
Screenshot from 2023-06-21 15-23-12

I'll refrain from posting screenshots of each error, but will share a summary:
- 1 out-of-memory error
- 3 other errors


package.json
```diff
@@ -14,18 +14,17 @@
         "format": "d2-style apply",
         "validate-commit": "d2-style check --staged",
         "validate-push": "yarn test",
-        "cy:open": "start-server-and-test 'yarn start' http://localhost:3000 'yarn cypress open --env networkMode=live'",
+        "cy:open": "start-server-and-test 'yarn start' http://localhost:3000 'yarn cypress open --e2e --browser electron --env networkMode=live'",
```
HendrikThePendric, line 17: Since Cypress now offers both e2e and component testing, the UI starts with a prompt which lets you decide which of these test types you want. By adding `--e2e` we skip this prompt.

By adding `--browser electron`, the tests start automatically in the integrated electron window. Running the tests in Chrome causes CORS issues, so causes all tests to fail. Seems we're stuck on Electron when using the Cypress UI, which shouldn't be a problem since it's also Chromium, and the more important Cypress runs _do_ use headless Chrome.

cypress/plugins/excludeByVersionTags.js
```diff
@@ -87,4 +88,15 @@ const getExcludedTags = (v) => {
...
+const excludeByVersionTags = (on, config) => {
+    const excludedTags = getExcludedTags(config.env.dhis2InstanceVersion)
+
+    console.log('instanceVersion', config.env.dhis2InstanceVersion)
+    console.log('tags to exclude from testing', excludedTags)
+
+    config.env.CYPRESS_EXCLUDE_TAGS = excludedTags.join(',')
+
+    on('file:preprocessor', tagify(config))
+}
```
HendrikThePendric, line 100: @jenniferarnesen: this logic was previously defined in `cypress/plugins/index.js` which is now removed. Most of what was previously there is now in `cypress.config.js`, but I didn't think this logic belonged in there. And since this is technically a plugin, moving it here seemed like the right thing to do.

cypress/integration/yourDimensions.cy.js
```diff
@@ -73,10 +73,18 @@ describe('event', () => {
...
+            // First is the actual container, second intersection detector wrapper
+            expect($elems).to.have.lengthOf(2)
+            const $container = $elems.first()
+            expect($container).to.have.class('container')
+            // Ensure intersection detector wrapper is excluded from options
+            const $options = $container.find('[data-test*="transfer-option"]')
+            // Ensure the only remaining option is the one matching the filter
+            expect($options).to.have.lengthOf(1)
+            expect($options.first()).to.have.text(filteredItemName)
+        })
```
HendrikThePendric, line 87: This test was failing due to a timing issue. The problem was that the reference to options was being created before the filter was being applied and DOM nodes present at the time would be detached. By placing the `should` directly after the container selector Cypress will keep retrying from there until it finds the expected children.

Perhaps there is a cleaner way 🤔 

But this does work, and we are still testing the same things as before.

cypress/plugins/excludeByVersionTags.js
jenniferarnesen: Fine with me as long as it still works :)

cypress/integration/yourDimensions.cy.js
```diff
@@ -73,10 +73,19 @@ describe('event', () => {
...
+            expect($elems).to.have.lengthOf(2)
+            const $container = $elems.first()
+            expect($container).to.have.class('container')
+            // Ensure the intersection detector wrapper is excluded from options
+            const $options = $container.find('[data-test*="transfer-option"]')
+            // Ensure the only option remains
+            expect($options).to.have.lengthOf(1)
+            // And it matches the filter
+            expect($options.first()).to.have.text(filteredItemName)
+        })
```
HendrikThePendric, line 88: This test was failing due to a timing issue. The problem was that the reference to options was being created before the filter was being applied and DOM nodes present at the time would be detached. By placing the `should` directly after the container selector, Cypress will keep retrying from there until it finds the expected children.

Perhaps there is a cleaner way 🤔 

But this does work, and we are still testing the same things as before.

cypress/support/e2e.js
```diff
@@ -0,0 +1,48 @@
...
+})
+
+// Log in before each test, if not already logged in
+beforeEach(() => {
+    cy.login({
+        name: Cypress.env('dhis2Username'),
+        password: Cypress.env('dhis2Password'),
+        server: Cypress.env('dhis2BaseUrl'),
+    })
+})
```
HendrikThePendric, line 48: This part of the PR unsettles me slightly. In `@dhis2/cypress-commands` we have a function called `enableAutlogin` which does something very similar to this, but when switching to use that the test suite became flaky.

Since `enableAutoLogin` sets `cacheAcrossSpecs` to `true` and the flakiness I was seeing was usually to do with tests being stuck at the login modal, the difference between `enableAutoLogin` and the login command here could explain things.

However, `@dhis2/cypress-commands` also exports a [`loginByApi`](https://github.com/dhis2/cli-utils-cypress/blob/master/packages/cypress-commands/src/commands/loginByApi.js) and a [`validateUserIsLoggedIn`](https://github.com/dhis2/cli-utils-cypress/blob/master/packages/cypress-commands/src/commands/validateUserIsLoggedIn.js) command. These commands do exactly the same thing as what I have above. So using this custom code or the commands from our lib _should_ produce the same results. But they didn't, and this both puzzles and concerns me.

PR #354, opened 2023-05-09
chore: disable video completely

HendrikThePendric:
A dummy PR to test the Cypress setting in CI which have been causing memory issues.

Should be closed when https://github.com/dhis2/line-listing-app/pull/345 is closed

The only reason this PR is not in DRAFT is because the e2e tests don't run otherwise.


HendrikThePendric:
Disabling video did not make any difference

PR #374, opened 2023-05-31
chore: cypress upgrade using custom login mechanism

HendrikThePendric:
Implements [DHIS2-15257](https://dhis2.atlassian.net/browse/DHIS2-15257)

---

### Key features

1. This PR enables us to use the latest Cypress version.
2. Retries failed tests once, so runs pass more often and flake stats are collected.
3. Only compresses and uploads video for failed tests, which makes the run faster, since this takes a long time.

---

### Description

The Cypress version we were on was trailing 4 major version behind the latest, but we were blocked from upgrading it by `@dhis2/cli-utils-cypress`. Now that that lib has been upgraded we can move to the latest Cypress. This latest version offers some bug fixes, a speed boost, and includes the `cy.session` API which should enable us to easily switch users in the test suite.

The Cypress upgrade itself was fairly straightforward but after the upgrade the test suite was no longer passing. In general I faced several categories of problems:
1. Some tests just needed to be tweaked a bit
2. Issues due to test-isolation config
3. Out of memory issues
4. Authentication issues

The issues in category 1 and 2 were OK to address because this was going wrong locally and consistently. But then the problems in category 3 and 4 were a different story. These were not always reproducible locally, but only happening on CI, and often I was seeing the same problem occur in different tests. As the test suite became more stable, it actually became harder to make sense out of it all, because only one out of multiple runs would fail, so sometimes I would think I had "fixed" something, only to realize later on that the problem was still there.

To keep things moving along I decided to take another, very slow but useful, quantitative approach. I consistently worked with two branches which were pretty much identical apart from 1 area. I then did about 20 runs for each branch and compared the results. Using this approach I eventually managed to get the test suite _fairly_ stable.

It also led me to chose the current implementation over the one in #345 which is using `cy.session`. It would have been nice use the new Cypress API to be idiomatic and to easily allow user switching in the future. However it turned out that:
1. This PR had fewer failures than PR#345: 1 out 21 runs with failures here VS 3 out of 21 failures in PR#345 
2. This PR never had the `Missing browserCriClient in connectToNewSpec` error, which indicates a memory issue. PR#345 had the out-of-memory issue 2 times.
3. The runs on this PR seem to complete faster, which could be explained by the fact that in this PR we only set a cookie before each test, while the `cy.session` implementation would actually login again before each test*. I should mention that I don't actually have any metrics about the run duration, I simply noticed that quite often the run on this PR was done, while the run on the other PR was still in progress, when I started them at the same time.

_[*] `cy.session()` has a `cacheAcrossSpecs` option, which would prevent logging in before each test, but when enabling that the out-of-memory issue appeared much more frequently_

#### Test isolation

This e2e test suite contains quite a few files with sequential tests. This which means that some test steps depend on a previous step. With test isolation switched on (the default setting in Cypress V12) each step (`it` block) will start off in a fresh window, and that breaks this kind
of test. So for these types of tests `testIsolation` was set to `false` at the `describe` block level.

---

### Known issues

-   The test suite is not completely stable
- We are not using `cy.session`

---

### Screenshots

cypress_12_screenshot

[DHIS2-15257]: https://dhis2.atlassian.net/browse/DHIS2-15257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric:
Here is a screen shot of the last 21 run results:
Screenshot from 2023-06-21 13-18-53

And here a screenshot of the only failed test we saw among those 21 run
Screenshot from 2023-06-21 14-23-11
s:

package.json
```diff
@@ -14,18 +14,17 @@
         "format": "d2-style apply",
         "validate-commit": "d2-style check --staged",
         "validate-push": "yarn test",
-        "cy:open": "start-server-and-test 'yarn start' http://localhost:3000 'yarn cypress open --env networkMode=live'",
+        "cy:open": "start-server-and-test 'yarn start' http://localhost:3000 'yarn cypress open --e2e --browser electron --env networkMode=live'",
```
HendrikThePendric, line 17: Since Cypress now offers both e2e and component testing, the UI starts with a prompt which lets you decide which of these test types you want. By adding `--e2e` we skip this prompt.

By adding `--browser electron`, the tests start automatically in the integrated electron window. Running the tests in Chrome causes CORS issues, so causes all tests to fail. Seems we're stuck on Electron when using the Cypress UI, which shouldn't be a problem since it's also Chromium, and the more important Cypress runs _do_ use headless Chrome.

cypress/integration/interpretations.cy.js
```diff
@@ -158,9 +158,17 @@ describe('interpretations', () => {
...
+            .contains('Update')
+            .should('have.class', 'loading')
+
         cy.getBySel('interpretations-list').contains(
             TEST_INTERPRETATION_TEXT_EDITED
         )
+
+        cy.getBySel('interpretations-list')
+            .contains('Update')
+            .should('not.exist')
```
HendrikThePendric, line 171: This test was passing on most DHIS2 Core versions, but not on 2.38. I managed to produce it locally and it turned out to be a timing issue. This is a sequential set of tests and the tests _after_ the  `the new interpretation can be edited` were expecting the saved interpretation to be present. However, when these tests started the "Update" button was still loading and the process had not been completed yet. So now I added two steps to verify the side panel goes into a loading state which eventually completes. This should make things more robust.

cypress/support/commands.js
```diff
@@ -49,3 +51,23 @@ Cypress.Commands.add(
...
+            cy.loginByApi(user)
+        },
+        {
+            validate: () => {
+                cy.validateUserIsLoggedIn(user)
+            },
+            cacheAcrossSpecs: false,
+        }
+    )
+})
```
HendrikThePendric, line 73: We could remove this, but I am still hopeful that at some point in the future logging in this way will actually work correctly. So it might be useful to keep this around.

cypress/support/e2e.js
```diff
@@ -0,0 +1,82 @@
...
+    const { name, value, ...options } = JSON.parse(Cypress.env(envVariableName))
+
+    localStorage.setItem(LOCAL_STORAGE_KEY, baseUrl)
+    cy.setCookie(name, value, options)
+
+    cy.getAllCookies().should((cookies) => {
+        expect(findSessionCookieForBaseUrl(baseUrl, cookies)).to.exist
+        expect(localStorage.getItem(LOCAL_STORAGE_KEY)).to.equal(baseUrl)
+    })
+})
```
HendrikThePendric, line 82: All this does is effectively mimic our old login behavior, which was based on persisting a cookie. Unfortunately Cypress has deprecated the methods to persist cookies in a normal way. So what I do here is this:
- Login before the run and persist a stringified version of the cookie in an environment variable
- Before each test read and parse the environment variable, and set the session cookie again.

.github/workflows/dhis2-verify-app.yml
```diff
@@ -27,7 +27,7 @@ jobs:
             - uses: actions/checkout@v3
             - uses: actions/setup-node@v3
               with:
-                  node-version: 18.x
+                  node-version: 16.x
```
jenniferarnesen, line 30: Could you just add a comment in the description about why node was downgraded?

.github/workflows/dhis2-verify-app.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     call-workflow-e2e-prod:
         if: "!contains(github.event.head_commit.message, '[skip ci]')"
         needs: [build, lint, test]
-        uses: dhis2/data-visualizer-app/.github/workflows/e2e-prod.yml@dev
+        uses: ./.github/workflows/e2e-prod.yml
```
jenniferarnesen, line 80: Are these workflows different now Could we port the changes to the "shared" copy in DV. Btw, the DV location is only temporary until we decide how to manage our shared workflows (e.g., the workflows repo)

cypress/integration/conditions/numericConditions.cy.js
```diff
@@ -233,7 +233,7 @@ describe('number conditions', () => {
     })
 })
 
-describe('integer', () => {
+describe('integer', { testIsolation: false }, () => {
```
jenniferarnesen, line 236: Could you add a comment in the PR description about why these tests needed testIsolation: false ?
That info could help when upgrading our other repos

jenniferarnesen (approved):

package.json
jenniferarnesen: I was not able to run the tests locally with cy:open using the electron browser. They do pass in chrome, however. We agreed during our slack huddle to remove the --browser config

.github/workflows/dhis2-verify-app.yml
HendrikThePendric: Good question. Yes there are some minor differences between what we have in DV and what we have here now. And I agree it'd be good to use the same workflow for LL and DV, or ideally for all 4 of our apps.

I'll bare this in mind when I upgrade DV.

.github/workflows/dhis2-verify-app.yml
HendrikThePendric: Thanks for pointing this out. I don't think I deliberately downgraded the node version, but I did want to make sure we were using the same version in all of the different workflow `.yaml` files.

I've now pushed a commit that tries v18 everywhere. If nothing crashes, I'll stick with that.

HendrikThePendric: It worked, so sticking with it.

cypress/integration/conditions/numericConditions.cy.js
HendrikThePendric: Good comment, thanks. This actually led me to do three things:
- Add an explanation to the PR description as you requested
- Add some comments to the codebase where it is used
- Review the usage of `testIsolation: false` and I think I added it in a few places where it _should not_ be required.

So I removed those occurrences in https://github.com/dhis2/line-listing-app/pull/374/commits/e05799240178bb82e9ac85b3e26251296b68ccac and if the tests for that commit pass, I will keep things like that.

PR #382, opened 2023-06-07
feat: persist sorting information in AO (DHIS2-14955)

edoardo:
Implements [DHIS2-14955](https://dhis2.atlassian.net/browse/DHIS2-14955)
Fixes [DHIS2-15248](https://dhis2.atlassian.net/browse/DHIS2-15248)

**Requires https://dhis2.atlassian.net/browse/DHIS2-14956**

---

### Key features

1. persist sorting details in LL AOs
2. apply sorting also to downloads

---

### Description

Moved the sorting field and direction state from the `Visualization` component to the `App`.
The sorting details are fetched directly in the visualization/current object, in this way the sorting is correctly applied to dashboard items too when first loaded.

~One regression tho, is that changing sorting does not work in dashboard.~
~The reason is that the `onDataSorted` callback is only passed to the plugin by the LL app.~
The above has been fixed by adding the `onDataSorted` callback to the `PluginWrapper` component used by dashboard.
 
---

### TODO

-   [x] figure out a solution for "temporarily" changing sorting in dashboard
-   [x] enable the new Cypress test when the [backend support](https://dhis2.atlassian.net/browse/DHIS2-14956
) is in place 
---

### Screenshots

Sorting applied and returned in the AO:
Screenshot 2023-07-11 at 10 29 12

Screenshot 2023-07-11 at 10 29 39

[DHIS2-14955]: https://dhis2.atlassian.net/browse/DHIS2-14955?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-15248]: https://dhis2.atlassian.net/browse/DHIS2-15248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cypress/integration/save.cy.js
```diff
@@ -52,6 +61,38 @@ describe('save', () => {
...
+        saveVisualization(AO_NAME)
+        expectAOTitleToContain(AO_NAME)
+        expectTableToBeVisible()
+
+        // apply sorting
+        getTableHeaderCells()
+            .find(`button[title*="${TEST_DIM_INTEGER}"]`)
+            .click()
+
+        expectTableToBeUpdated()
```
martinkrulltott, line 78: add a step that verifies that the actual values in the table are sorted? e.g. change "2, 3, 1" to "1, 2, 3".

Also add this verification step to when the AO has been loaded again below.

cypress/integration/save.cy.js
```diff
@@ -52,6 +61,38 @@ describe('save', () => {
...
+            .its('request.body')
+            .should('have.property', 'sorting')
+
+        cy.wait('@loadAO')
+            .its('response.body')
+            .should('have.property', 'sorting')
+
+        deleteVisualization()
+    })
+
```
martinkrulltott, line 95: consider adding a Cypress test for what happens if you try to save a sorting prop that’s not part of the visualisation

src/components/Visualization/Visualization.js
```diff
@@ -108,24 +108,50 @@ export const Visualization = ({
...
 
     const visualization = useMemo(() => AO && transformVisualization(AO), [AO])
 
     const visualizationRef = useRef(visualization)
 
+    const getSorting = (visualization) => {
+        if (
+            visualization &&
+            visualization.sorting &&
+            visualization.sorting.length
```
martinkrulltott, line 124: refactor to `if ( visualization?.sorting?.length )`?

martinkrulltott: or even just skip the `if` and do a conditional return like `cost getSorting = (visualization) => visualization?.sorting?.length ? { foo } : { bar }`

edoardo: Refactored.

cypress/integration/save.cy.js
edoardo: There are tests checking the sorting on the values in `table.cy.js`.

But I've added a test that checks the sorting after loading the AO just saved.

cypress/integration/save.cy.js
edoardo: I don't think that is possible, if it's not part of the visualisation is not saved.
Or do you mean simulate a POST request with "tempered" payload?
(Is this even a valid pattern for Cypress testing?)

edoardo: Added a test for verifying that sorting data is removed when a column with applied sorting is removed from layout.

cypress/integration/save.cy.js
```diff
@@ -90,6 +94,59 @@ describe('save', () => {
...
+            .then(parseInt)
+            .then(($cell0Value) =>
+                getTableRows()
+                    .eq(1)
+                    .find('td')
+                    .eq(1)
+                    .invoke('text')
+                    .then(parseInt)
+                    .then(($cell1Value) =>
+                        expect($cell0Value).to.be.lessThan($cell1Value)
```
martinkrulltott, line 113: maybe technically correct, but why not just check for specific values? like expect cell0 to be 15 and cell1 to be 20 etc. That's usually the preferred reliable way.

martinkrulltott (approved):

cypress/integration/save.cy.js
edoardo: Refactored.

cypress/integration/save.cy.js
```diff
@@ -90,6 +94,59 @@ describe('save', () => {
...
+        expectTableToBeUpdated()
+
+        cy.intercept('POST', /eventVisualizations\?/).as('saveAO')
+
+        saveVisualizationAs(AO_NAME)
+
+        cy.wait('@saveAO')
+            .its('request.body')
+            .should('not.have.property', 'sorting')
+
```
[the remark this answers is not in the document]
edoardo: Added `expectTableToBeVisible`.

jenniferarnesen:
LGTM

PR #388, opened 2023-06-15
fix: avoid crash when missing metadata for stage-element ids (DHIS2-15423)

janhenrikoverland:
When loading a saved line list with multistage data elements we don't have metadata on the "stageId.elementId" format in the loaded visualization object. This causes a crash in the `SelectedDimensionsContext.js` file.

This is a hotfix and simply avoids the destructuring of a non existing object.

The proper fix would be to construct the correct metadata and add it to the metadata store based on the saved line list. With this fix we are not able to restore the dimensions in the ui if the analytics request fails.


HendrikThePendric (approved):
Could also use `??` instead of `||` but this is functionally equivalent. 👍

PR #396, opened 2023-06-21
feat: support new dashboard plugin architecture

edoardo:
Part of the fix for: [DHIS2-17268](https://dhis2.atlassian.net/browse/DHIS2-17268)

**Requires https://github.com/dhis2/analytics/pull/1672**

---

### Key features

1. use shared dashboard plugin wrapper

---

### Description

Generic components for plugins are available in `app-platform` and `app-runtime`.
Analytics apps are currently having the same components replicated in each repo, a shared dashboard plugin wrapper has been added in `analytics`.

---

### Screenshots
Screenshot 2023-06-21 at 16 25 33

[DHIS2-17268]: https://dhis2.atlassian.net/browse/DHIS2-17268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Visualization/Visualization.js
```diff
@@ -178,27 +187,29 @@ export const Visualization = ({
...
+            console.log('resize container', node)
             if (node === null) {
                 return
             }
 
             containerRef.current = node
             sizeObserver.observe(node)
 
-            return sizeObserver.disconnect
+            return () => sizeObserver.unobserve(node)
```
HendrikThePendric, line 198: I'm not sure if this change actually makes a difference in practice because both elements probably unmount at the same time and calling `.disconnect()` multiple times is probably not a big deal. Having said that, I'd still be in favour of keeping it: in the functions we call observe on a single node, so in the cleanup we should stick to just unobserve that particular node.

src/components/Visualization/Visualization.js
```diff
@@ -178,27 +187,29 @@ export const Visualization = ({
...
         },
         [sizeObserver]
     )
 
     const observeVisualizationContainerRef = useCallback(
         (node) => {
+            console.log('resize vis container', node)
```
HendrikThePendric, line 205: Can also be removed
```suggestion
```

HendrikThePendric (approved):
LGTM 🚀

src/components/Visualization/Visualization.js
edoardo: Yes, I left it there, but it didn't seem to make a difference.
What made the difference was to add the `requestAnimationFrame`.

src/components/Dialogs/PeriodDimension/PeriodDimension.js
```diff
@@ -145,7 +113,7 @@ export const PeriodDimension = ({ dimension, onClose }) => {
                 if (isStartEndDate(item.id)) {
                     acc.metadata[item.id] = {
                         id: item.id,
-                        name: formatStartEndDate(item.id),
+                        name: item.id.replace('_', ' - '),
```
HendrikThePendric, line 116: This part is a little bit confusing to me. It seems to me that for these `startEndDates`, we don't actually use the `name` property, but we always use the ID (string) and then parse it. So why do we need to do a string substitution on the `name`?

HendrikThePendric (commented):
The changes look good, and probably everything makes sense, but there was one part that confused me a bit.

src/components/Dialogs/PeriodDimension/PeriodDimension.js
HendrikThePendric: As discussed in a call: this is to keep the metadata consistent. I'll resolve the conversation and approve.

HendrikThePendric (approved):

PR #402, opened 2023-06-26
fix: use local copy of e2e-dev in nightly github workflow

HendrikThePendric:
Implements a fix for the nightly job, which was broken after https://github.com/dhis2/line-listing-app/pull/374 was merged.

---

### Key features

1. Will (hopefully) avoid the Cypress out-of-memory issue on CI

---

### Description

When working on https://github.com/dhis2/line-listing-app/pull/374 I tweaked a lot of things to reduce the likelyhood of seeing the `Missing browserCriClient in connectToNewSpec` error occur during the Cypress e2e test run.

One of the things I did was to start working with a local copy of the `e2e-prod` GitHub workflow. I did this because the shared version of this workflow from data-visualizer is using a pre-baked Docker container in which the error was occuring very frequently. Not using that container made the error occur less frequently.

Whilst working on https://github.com/dhis2/line-listing-app/pull/374 I never really paid much attention to the nightly workflow, and upon inspection this is also using a shared workflow from data-visualizer, and this workflow is also using a docker container.

So what I've done in this PR is:
- Create a local copy of e2e-dev which *doesn't* use a container.
- Switched to using that local workflow in the nighly workflow.

### DIFF
--- a/.github/workflows/e2e-dev.yml
+++ b/.github/workflows/e2e-dev.yml
@@ -1,0 +1,73 @@
+name: 'e2e-dev'
+
+on:
+    workflow_call:
+        secrets:
+            baseurl:
+                required: true
+            username:
+                required: true
+            password:
+                required: true
+            recordkey:
+                required: true
+
+concurrency:
+    group: e2e-dev-${{ github.workflow}}-${{ github.ref }}
+    cancel-in-progress: true
+
+defaults:
+    run:
+        shell: bash
+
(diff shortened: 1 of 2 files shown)


jenniferarnesen (approved):

HendrikThePendric:
FYI: When squash merging I changed the commit prefix from `fix` to `chore`, because this should not trigger the release of a new version.

PR #418, opened 2023-07-19
chore: spin up local instances

HendrikThePendric:
WIP.... Very much a WIP
But not in draft as I need to run all the github actions here....


jenniferarnesen:
Taking a different approach

PR #426, opened 2023-09-04
fix: handle options with non-unique codes across optionsets (DHIS2-15771)

edoardo:
This requires a backend fix.

Fixes [DHIS2-15771](https://dhis2.atlassian.net/browse/DHIS2-15771)
Fixes [DHIS2-16181](https://dhis2.atlassian.net/browse/DHIS2-16181)

**Requires [DHIS2-15775](https://dhis2.atlassian.net/browse/DHIS2-15775)**

---

### Key features

1. lookup option set option labels correctly in the table rows
2. show the correct name in the layout chip tooltips
3. refactor how the metadata for option sets is stored in Redux

---

### Description

Options for different option sets can use the same `code`, which is what is returned in the rows in the analytics response.
The previous way of looking up the label to display in the table simply tried to find a metadata object whom `code` matches the value for the option set in the response's rows.
This cause to display the wrong label when different option sets are displayed at the same time and happen to use the same `code`.

Refer to the screenshots below.
The 3 columns are 3 different option sets, but they all use the same `code` (**1**) for their option.
The first matching `code` found in metadata items (in the analytics response) was always used regardless of the option set.

The fix is to look in metadata first for the `optionSet`, and find the matching option by `code` within the listed `options`.
Then the option's id is used to lookup the option itself in metadata and from there the `name` is extracted and used in the table.

The same logic needs to be applied when looking up the name to display in the layout chip tooltip for dimensions with option set where there is a specific selection (not all values).

We also need to add the metadata for the options when loading a saved visualization.
The reason here is that the table might not have all the values there are selected for the option set, and in that case the metadata from the analytics response is partial.

While working on this I realised that we don't really need to store the metadata for the single options of an option set.
Since we need them in the `options` object in the option set metadata anyway for the lookup, we can have the `name` in there as well and skip one lookup.

---

### TODO

-   [x] Cypress tests
-   [x] Update docs?
-   [x] Manual testing

---

### Known issues

-   [x] a Cypress test in `table.cy.js` is now failing due to this change. This is not a regression, it should be working once the backend fix is available.

---

### Screenshots

Table before:
Screenshot 2023-09-04 at 14 03 38

Table after:
Screenshot 2023-09-04 at 13 57 48

Tooltip before:
Notice the names for the selected options are wrong when a different dimension with option set has also a selection and the `code` used is the same:
Screenshot 2023-11-23 at 11 19 45

Screenshot 2023-11-23 at 11 19 51

In this example, this selection was made first:
Screenshot 2023-11-23 at 11 19 25

Tooltip after:
Notice the Pain medication dimension has 5 options selected, but the tooltip for pre-eclampsia is correct:
Screenshot 2023-11-23 at 11 23 08

[DHIS2-15771]: https://dhis2.atlassian.net/browse/DHIS2-15771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DHIS2-15773]: https://dhis2.atlassian.net/browse/DHIS2-15773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-15775]: https://dhis2.atlassian.net/browse/DHIS2-15775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-16181]: https://dhis2.atlassian.net/browse/DHIS2-16181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -44,8 +44,21 @@ const excludedDimensions = [
     DIMENSION_ID_LAST_UPDATED_BY,
 ]
 
-const findOptionSetItem = (code, metaDataItems) =>
-    Object.values(metaDataItems).find((item) => item.code === code)
+const findOptionSetItem = (optionSetId, code, metaDataItems) => {
+    const optionSetMetaData = metaDataItems[optionSetId]
+
+    if (optionSetMetaData) {
+        const optionId = optionSetMetaData.options.find(
+            (option) => option.code === code
+        )?.id
+
+        if (optionId) {
+            return metaDataItems[optionId]
+        }
+    }
+
+    return undefined
+}
 
 const formatRowValue = (rowValue, header, metaDataItems) => {
     switch (header.valueType) {
@@ -58,7 +71,8 @@ const formatRowValue = (rowValue, header, metaDataItems) => {
             }
             if (header.optionSet) {
                 return (
-                    findOptionSetItem(rowValue, metaDataItems)?.name || rowValue
+                    findOptionSetItem(header.optionSet, rowValue, metaDataItems)
+                        ?.name || rowValue
                 )
             }
             return metaDataItems[rowValue]?.name || rowValue
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -45,7 +45,7 @@ const excludedDimensions = [
 ]
 
 const findOptionSetItem = (optionSetId, code, metaDataItems) => {
-    const optionSetMetaData = metaDataItems[optionSetId]
+    const optionSetMetaData = metaDataItems?.[optionSetId]
 
     if (optionSetMetaData) {
         const optionId = optionSetMetaData.options.find(
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -50,7 +50,7 @@ const findOptionSetItem = (optionSetId, code, metaDataItems) => {
     if (optionSetMetaData) {
         const optionId = optionSetMetaData.options.find(
             (option) => option.code === code
-        )?.id
+        )?.uid
 
         if (optionId) {
             return metaDataItems[optionId]
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -52,9 +52,7 @@ const findOptionSetItem = (optionSetId, code, metaDataItems) => {
             (option) => option.code === code
         )?.uid
 
-        if (optionId) {
-            return metaDataItems[optionId]
-        }
+        return metaDataItems[optionId]
     }
 
     return undefined
--- a/src/components/Visualization/useAnalyticsData.js
+++ b/src/components/Visualization/useAnalyticsData.js
@@ -44,7 +44,7 @@ const excludedDimensions = [
     DIMENSION_ID_LAST_UPDATED_BY,
 ]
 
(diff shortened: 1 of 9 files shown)


src/components/Visualization/useAnalyticsData.js
```diff
@@ -43,8 +43,21 @@ const excludedDimensions = [
     DIMENSION_ID_LAST_UPDATED_BY,
 ]
 
-const findOptionSetItem = (code, metaDataItems) =>
-    Object.values(metaDataItems).find((item) => item.code === code)
+const findOptionSetItem = (optionSetId, code, metaDataItems) => {
+    const optionSetMetaData = metaDataItems[optionSetId]
```
janhenrikoverland, line 47: The only way I could break this function was by passing null or undefined for `metaDataItems` (or leave it out). Maybe do
`const optionSetMetaData = metaDataItems?.[optionSetId]`
to avoid that.

edoardo: Done.

src/components/Visualization/useAnalyticsData.js
```diff
@@ -43,8 +43,21 @@ const excludedDimensions = [
...
+        const optionId = optionSetMetaData.options.find(
+            (option) => option.code === code
+        )?.uid
+
+        if (optionId) {
+            return metaDataItems[optionId]
+        }
+    }
+
+    return undefined
```
martinkrulltott, line 59: Given how this fn is used I don't think you need to return anything here?
But if you want to make it explicit, I think the correct return value would be `null`.

src/components/Visualization/useAnalyticsData.js
```diff
@@ -43,8 +43,21 @@ const excludedDimensions = [
...
-    Object.values(metaDataItems).find((item) => item.code === code)
+const findOptionSetItem = (optionSetId, code, metaDataItems) => {
+    const optionSetMetaData = metaDataItems?.[optionSetId]
+
+    if (optionSetMetaData) {
+        const optionId = optionSetMetaData.options.find(
+            (option) => option.code === code
+        )?.uid
+
+        if (optionId) {
```
martinkrulltott, line 54: Unnecessary check? `metaDataItems[optionId]` will return undefined if not found anyway, so no need to wrap it?

martinkrulltott (commented):
`findOptionSetItem` could be written as a single line, but I see that splitting it up increases readability. However, I'd argue that some parts are a bit unnecessary/too explicit right now, which actually decreases readability.

martinkrulltott:
Needs Cypress tests

src/components/Visualization/useAnalyticsData.js
edoardo: My reasoning was to return the same value the function can return above where the `find` is used.

src/components/Visualization/useAnalyticsData.js
edoardo: I removed the condition.

edoardo:
> Needs Cypress tests

Added.

martinkrulltott (approved):

janhenrikoverland (approved):

PR #428, opened 2023-09-04
fix: allow opening interpretations for AOs without a time dimension (DHIS2-15781)

HendrikThePendric:
Implements [DHIS2-15781](https://dhis2.atlassian.net/browse/DHIS2-15781)

**Requires https://github.com/dhis2/analytics/pull/1571**

---

### Key features

1. Allow opening interpretations for AOs without a time dimension

---

### Description

When opening a visualization in the interpretations modal, we pass a `relativePeriodDate` to ensure we are only showing data that was avaiable when the interpretation was created. However, some visualizations do not include a time dimension and for these visualizations adding the `relativePeriodDate` parameter does not make sense and sometimes also casues an error.

So this PR adds a `isAoWithTimeDimension` function which can be used to check if the visualization (AO object) includes a valid time dimension and only adds the `relativePeriodDate` param if that functions returns `true`.

I'd say the thinking behind this makes sense, but I am not at all sure if `isAoWithTimeDimension` is correctly catching all visualizations with a valid time dimension. When I was testing things locally the issue seemed solved, but I have no idea if the logic behind it is sound and/or complete....

---

### Consideration

After creating this PR something occurred to me which might make us want to close this PR and turn this into a backend issue:
- As I understand it, the whole point of providing the `relativePeriodDate` parameter is that the visualization does not include any data which was added after the interpretation was written, so that the user reading the interpretation is looking at exactly the same thing as the person who wrote it.
- The fix I have produced here _omits_ the `relativePeriodDate` parameter, and the consequence of this is that the person reading the interpretation at a later date may be looking at something quite different than the person who originally wrote the interpretation. So while my fix "solves" the issue from a technical perspective (no error response, and data is being displayed), it comes with some considerable disadvantages as well.
- And assuming a data-point also has a created-at/updated-at field in the DB, it seems that this issue can simply be addressed by excluding data which was created (or perhaps updated?) after the `relativePeriodDate`. This seems like a better solution to me, and this would make it a backend issue.

---

### Decision

@janhenrikoverland, @cooper-joe & @maikelarabori discussed the consideration above and it was decided that we would stick with the current implementation with the addition of some warning messages. We show one when writing the interpretation and one when viewing one. The addition of this warning when writing the interpretation had to be done in the analytics repo.

---

### Screenshots

Screenshot 2023-09-25 at 16 05 32
Screenshot 2023-09-25 at 16 05 11

[DHIS2-15781]: https://dhis2.atlassian.net/browse/DHIS2-15781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/timeDimensions.js
```diff
@@ -123,3 +124,11 @@ export const getDisabledTimeDimensions = (inputType, program, stage) => {
...
     }
 }
+
+export const isAoWithTimeDimension = ({ columns }) =>
+    columns.some(
+        ({ dimension, items }) =>
+            DIMENSION_IDS_TIME.has(dimension) &&
+            Array.isArray(items) &&
+            items.length > 0
+    )
```
HendrikThePendric, line 134: Please triple check this makes sense

janhenrikoverland: We have tools for this in the analytics lib. I'll send you a link.

HendrikThePendric: @jan I rewritten this and did use helpers from analytics, but not exactly the one you pointed me to....
```js
export const isAoWithTimeDimension = (ao) =>
    layoutGetAllDimensions(ao).some(
        ({ dimensionType, dimension, items }) =>
            (dimensionType === DIMENSION_TYPE_PERIOD ||
                DIMENSION_IDS_TIME.has(dimension)) &&
            Array.isArray(items) &&
            items.length > 0
    )
```

HendrikThePendric:
To test:
- I have created an environment which includes both the changes from this PR as well as the related changes in analytics
- [Here's a link](https://dev.im.dhis2.org/analytics-v40-dhis2-14827/api/apps/line-listing/index.html#/AFjkDs7acBh) to a line list _with_ a time dimension, which shouldn't show the warnings:
- And [here's one](https://dev.im.dhis2.org/analytics-v40-dhis2-14827/api/apps/line-listing/index.html#/HC1qO3gRNPB) _without_ a time dimension, which should show the warnings.
- Warnings are in two places:
    - Below the rich text editor when creating the interpretation
    - Above the visualization in the interpretations modal

janhenrikoverland (approved):

janhenrikoverland (approved):

HendrikThePendric:
Closing in favour of https://github.com/dhis2/line-listing-app/pull/447

PR #434, opened 2023-09-11
fix: always attach resize observer if node is found

HendrikThePendric:
Fixes a regression in [DHIS2-15465](https://dhis2.atlassian.net/browse/DHIS2-15465)

---

### Key features

1. This makes up for a shortcoming in the fix implemented in https://github.com/dhis2/line-listing-app/pull/431. The pagination now has the correct dimensions in a modal too.

---

### Description

The following problem would occur when opening an interpretation from the main canvas, but not when refreshing the page with an interpretations ID in the URL:
- While opening the `node.clientWidth` would temporarily be `0` which would prevent the resize observer from being attached.
- And without the resize observer, the state update would not be triggered once the container does get a width.
- The check on `node.clientWidth === 0` was added to prevent a React error when the modal closes. In this case the `clientWidth` is `0` and we want to prevent the state update to fire becasue at this point the component is unmounting. However, the check was in the wrong place: we want to prevent the state update, but we do not want to prevent the resize observer being attached to the node.

Since this is a fix for a regression of a PR which was merged into master by accident, we should probably organise a quick test session for this PR and then merge it into master ASAP.

---

### TODO

-   [ ] Needs to be checked in a plugin view too. Modal and main canvas have been checked.

---

### Screenshots

_supporting text_


src/components/Visualization/Visualization.js
```diff
@@ -301,6 +304,7 @@ export const Visualization = ({
     )
 
     const isInModal = !!filters?.relativePeriodDate
+    console.log('render', paginationMaxWidth)
```
martinkrulltott, line 307: intentional or for testing purposes only?

martinkrulltott (approved):

martinkrulltott (approved):

src/components/Visualization/Visualization.js
```diff
@@ -64,6 +64,8 @@ import {
 export const DEFAULT_SORT_DIRECTION = 'asc'
 export const FIRST_PAGE = 1
 export const PAGE_SIZE = 100
+// min width of "Rows per page" select + 2 * padding
```
martinkrulltott, line 67: what happens when a different language is used?

martinkrulltott (approved):

src/components/Visualization/Visualization.js
HendrikThePendric: Shoot. I hadn't considered that. I just tested this and things do break if I add text to the "Rows per page" pre-fix on the select, so well spotted. 

Since we can't know the `width` of this select upfront, the only solution would be to measure it. And this would involve adding a `ref` to that select, which we can't currently as it's part of the `Pagination` component in @dhis2/ui. However, adding a prop for it there would also be quite odd, I guess we'd have to call it `pagesizeSelectRef` and then (for consistency's sake) we would probably want to add these types of props for all the sub-components. Or alternatively/preferably, we would expose all the individual sub-components and allow them to accept a ref, and probably then we need to expose some functions which make up the internal logic of the component.

I'm not saying this is impossible. But I am wondering if such a high-effort solution would be worth it for what is effectively a bit of an edge case: this problem only occurs if the complete table has a smaller width than the "Rows per page" select.

I think I'd be in favour of keeping a hard coded value, until we receive any actual bug reports for this. Perhaps we should add a bit of a safety margin and increase it by 30% to accommodate for the fact that English is quite a concise language?

PR #453, opened 2023-10-30
fix: your- and program-dimensions list lazy loading

HendrikThePendric:
Implements [DHIS2-16083](https://dhis2.atlassian.net/browse/DHIS2-16083)

---

### Key features

1. Fixes lazy loading of the "your dimensions" list

---

### Description

Three unrelated issues were preventing this list from lazy loading properly:
1. Due to a CSS issue the scroll-box container was actually the scrolling element instead of the scrollbox itself. This caused three issues:
    1. Since the scrollbox was growing to the full list height, the list was "lazy loading" all of its pages in one go, because the end of the list was always visible.
    2. Scrolling down to the bottom would not trigger anything to happen.
    3. The search input would scroll out of view when scrolling down
2. When computing the next page we were incrementing from a wrong property causing the next page to be `NaN`
3. The `DimensionList` component assumes that when additional items are being _lazy loaded_ `loading` is `false` but `fetching` is `true`. Instead both variables were `true` and as a result the "loading initial list UI" was returned instead of the "loading additional items UI". The `useYourDimensions` hook has been adjusted so that it behaves as the `DimensionList` component expects.

---

### Test instructions

To reproduce this issue we need to have a lot of organisation-unit-group-sets in the test instance. I have a attached a JSON file which can be used to quickly add them using the import/export-app. I'll provide some detailed instructions below:
1. Download [organisationUnitGroupSets.json](https://github.com/dhis2/line-listing-app/files/13204567/organisationUnitGroupSets.json) to your computer
2. On the instance you'd like to use for testing, visit the import/export-app and go to the "Metadata import section"
4. Select the JSON file, leave all the default settings and hit the "Start import" button.

---

### Screenshots

your_dimensions_lazy_loading

[DHIS2-16083]: https://dhis2.atlassian.net/browse/DHIS2-16083?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cypress/integration/yourDimensions.cy.js
+++ b/cypress/integration/yourDimensions.cy.js
@@ -1,5 +1,6 @@
 import { DIMENSION_ID_EVENT_DATE } from '../../src/modules/dimensionConstants.js'
 import { E2E_PROGRAM, TEST_REL_PE_LAST_YEAR } from '../data/index.js'
+import yourDimensionsFixture from '../fixtures/yourDimensionsLazyLoading.json'
 import { typeInput } from '../helpers/common.js'
 import {
     openProgramDimensionsSidebar,
@@ -25,11 +26,11 @@ const trackerProgram = E2E_PROGRAM
 const periodLabel = trackerProgram[DIMENSION_ID_EVENT_DATE]
 
 describe('event', () => {
-    it('Your dimensions can be used and filtered by', () => {
-        const dimensionName = 'Facility Type'
-        const filteredOutItemName = 'MCHP'
-        const filteredItemName = 'CHC'
+    const dimensionName = 'Facility Type'
+    const filteredOutItemName = 'MCHP'
+    const filteredItemName = 'CHC'
 
+    const openYourDimensionsPanel = () => {
         goToStartPage()
 
         selectEventWithProgram(trackerProgram)
@@ -43,6 +44,10 @@ describe('event', () => {
 
         // open the your dimensions sidebar
         cy.getBySel('main-sidebar').contains('Your dimensions').click()
+    }
+
+    it('Your dimensions can be used and filtered by', () => {
+        openYourDimensionsPanel()
 
         cy.getBySel('your-dimensions-list').contains(
             dimensionName,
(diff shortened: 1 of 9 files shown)


src/components/MainSidebar/DimensionsList/DimensionsList.module.css
```diff
@@ -1,4 +1,5 @@
 .scrollbox {
+    height: calc(100% - 57px);
```
HendrikThePendric, line 2: This is not ideal since it depends on the height of a different component (the search box). But I don't know of a better way to do this.

martinkrulltott (commented):
The fix LGTM (although not ideal as mentioned in your comment). 
Some Cypress tests with `intercept` would be great until we've figured out the data in the database situation.

src/components/MainSidebar/DimensionsList/DimensionsList.module.css
martinkrulltott: Replace this by adding `display: flex; flex-direction: column;` to the parent?
image

HendrikThePendric: Awesome, works well.

cypress/integration/yourDimensions.cy.js
```diff
@@ -118,4 +123,60 @@ describe('event', () => {
             `${getPreviousYearStr()}-12-11`,
         ])
     })
+    it('Your dimension list lazy loads', () => {
+        const getList = () => cy.getBySel('your-dimensions-list')
+        const getListChildren = () => getList().find('div[role="button"]')
+        const shouldLoadNextPage = (nextPage, nextListLenght) => {
```
martinkrulltott, line 129:
```suggestion
        const shouldLoadNextPage = (nextPage, nextListLength) => {
```

martinkrulltott (approved):

PR #458, opened 2023-11-21
fix: always show unformatted date

HendrikThePendric:
No JIRA ticket avaiable

---

### Key features

1. Consistent custom date range strings in tooltip

---

### Description

Before, when setting a custom date range / defining a start and end date, after the first edit you would see a "pretty date" in the tooltip, but after each subsequent error, an unformatted date would show in that same tooltip. This PR makes the tooltip behaviour consistent by always showing an unformastted date. See [this Slack thread](https://dhis2.slack.com/archives/G3TL0J145/p1700493355195899) for a full discussion.

|  | Before fix | After fix |
| ------------- | ------------- | ------------- |
| After first selection  | inital_pre_fix |  initial_post_fix |
| After subsequent selections  |  subsequent | subsequent  |

### DIFF
--- a/cypress/integration/period.cy.js
+++ b/cypress/integration/period.cy.js
@@ -100,7 +100,7 @@ describe('period dimension', { testIsolation: false }, () => {
             .trigger('mouseover')
 
         cy.getBySelLike('tooltip-content').contains(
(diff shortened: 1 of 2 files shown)


jenniferarnesen:
Superceded

PR #472, opened 2023-12-13
chore: only upload videos for failing cypress tests

HendrikThePendric:
Not in JIRA

---

### Key features

1. Stop uploading videos of passing tests

### How to verify this is working correctly

1. The test suite should still pass
2. The test suite should not crash because it is trying to delete a file
3. When reviewing the cypress test run log we should see that no video was upload (unless there was a failing test)


martinkrulltott (approved):

HendrikThePendric:
Screenshot 2023-12-13 at 13 18 46

PR #477, opened 2024-01-09
test: dynamic offset for years

martinkrulltott:
No JIRA

Fixes issues related to the yearly bump in the database.
Basically turns things like `1991` to "this year minus 33" (i.e. 2024-33=1991) for data that was unexpectedly bumped in the demo db.

### DIFF
--- a/cypress/integration/conditions/dateConditions.cy.js
+++ b/cypress/integration/conditions/dateConditions.cy.js
@@ -23,6 +23,7 @@ import {
     getPreviousYearStr,
     unselectAllPeriods,
     selectFixedPeriod,
+    getOffsetYearStr,
 } from '../../helpers/period.js'
 import { goToStartPage } from '../../helpers/startScreen.js'
 import {
@@ -78,7 +79,7 @@ describe('date conditions (Date)', { testIsolation: false }, () => {
     })
 
     it('exactly', () => {
-        const TEST_DATE = '1991-05-21'
+        const TEST_DATE = `${getOffsetYearStr(32)}-05-21`
 
         addConditions([
             {
@@ -95,7 +96,7 @@ describe('date conditions (Date)', { testIsolation: false }, () => {
     })
 
     it('is not', () => {
-        const TEST_DATE = '1991-05-20'
+        const TEST_DATE = `${getOffsetYearStr(32)}-05-20`
 
         addConditions([
             {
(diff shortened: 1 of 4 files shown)


cypress/integration/timeDimensions.cy.js
```diff
@@ -27,7 +27,18 @@ const assertTimeDimension = (dimension) => {
         selectEventWithProgram(trackerProgram)
         openProgramDimensionsSidebar()
         const label = trackerProgram[dimension.id]
-        selectRelativePeriod({ label, period: TEST_REL_PE_THIS_YEAR })
+        if (dimension.id === DIMENSION_ID_LAST_UPDATED) {
+            // last updated doesn't seem to get bumped yearly in the SL DB so we need to select a fixed period
+            selectFixedPeriod({
+                label,
+                period: {
+                    year: '2023',
```
jenniferarnesen, line 35: Could we use current year from the Date object instead? If we did that, would it mean that we don't need to fix this every January?

cypress/integration/conditions/dateConditions.cy.js
```diff
@@ -78,7 +79,7 @@ describe('date conditions (Date)', { testIsolation: false }, () => {
     })
 
     it('exactly', () => {
-        const TEST_DATE = '1991-05-21'
+        const TEST_DATE = `${getOffsetYearStr(32)}-05-21`
```
jenniferarnesen, line 82: Could 32 and 33 be put into a const?

cypress/integration/table.cy.js
```diff
@@ -94,7 +95,7 @@ const programDimensions = [
 const programDataDimensions = [
     { label: TEST_DIM_AGE, value: '1991-01-01' },
```
jenniferarnesen, line 96: Does this '1991' value change? (same question for line 99)

martinkrulltott: No for some reason the `AGE` and `DATETIME` remained static in the yearly bump, so I left them as is.

cypress/integration/timeDimensions.cy.js
martinkrulltott: As commented in the code above, `Last updated` doesn't actually get bumped, so it will statically remain `2023` (unless Phil changes the script). That's why I had to make a special case for it (all other time dimensions are using `TEST_REL_PE_THIS_YEAR` because they actually get bumped)

jenniferarnesen (approved):

PR #480, opened 2024-01-12
feat: prefix time dimension ids with program id when tracked entity output type

jenniferarnesen:
Implements https://dhis2.atlassian.net/browse/DHIS2-16023

### Description

Time dimensions need to be prefixed with the programId when using TE output type. The following works:
- adding time dimensions from different programs to layout - they get suffixed with the program name
- analytics query works - data shows
- saving LL with time dimensions works
- opening a saved LL with time dimensions - layout is empty but table shows correctly (metadata issue to be solved with backend changes)

The failing cypress tests are the same as the target branch

image

image

### DIFF
--- a/src/components/Dialogs/PeriodDimension/PeriodDimension.js
+++ b/src/components/Dialogs/PeriodDimension/PeriodDimension.js
@@ -132,7 +132,6 @@ export const PeriodDimension = ({ dimension, onClose }) => {
             (acc, item) => {
                 const itemId = formatDimensionId({
                     dimensionId: item.id,
-                    programStageId: null,
                     programId,
                     outputType: inputType,
                 })
--- a/src/components/Dialogs/PeriodDimension/PeriodDimension.js
+++ b/src/components/Dialogs/PeriodDimension/PeriodDimension.js
@@ -25,7 +25,10 @@ import {
     SYSTEM_SETTINGS_HIDE_BIMONTHLY_PERIODS,
 } from '../../../modules/systemSettings.js'
 import { USER_SETTINGS_UI_LOCALE } from '../../../modules/userSettings.js'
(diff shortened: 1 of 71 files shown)


src/modules/metadata.js
```diff
@@ -70,10 +74,18 @@ export const getProgramAsMetadata = (program) => ({
...
-export const getDynamicTimeDimensionsMetadata = (program, stage) => ({
+export const getDynamicTimeDimensionsMetadata = (
+    program,
+    stage,
+    inputType
+) => ({
     ...Object.values(getTimeDimensions()).reduce((acc, dimension) => {
-        acc[dimension.id] = {
-            id: dimension.id,
+        const dimensionId = getTimeDimensionMetadataId(
```
martinkrulltott, line 83: It might be possible to replace `getTimeDimensionMetadataId` with the existing `formatDimensionId` like:

`formatDimensionId({ dimensionId: dimension.id, programId: program?.id, outputType: inputType })`

src/components/Dialogs/PeriodDimension/PeriodDimension.js
```diff
@@ -118,18 +120,23 @@ export const PeriodDimension = ({ dimension, onClose }) => {
...
 
+    const inputType = useSelector(sGetUiInputType)
+
+    const { programId } = extractDimensionIdParts(dimension.id, inputType)
+
     const updatePeriodDimensionItems = (items) => {
         const { uiItems, metadata } = items.reduce(
             (acc, item) => {
-                acc.uiItems.push(item.id)
+                const itemId = programId ? `${programId}.${item.id}` : item.id
```
martinkrulltott, line 130: This should use the utils instead to make sure program id is only added when applicable (i.e. when input === tracked entity)

martinkrulltott (approved):

PR #483, opened 2024-01-15
fix: chips having same dimension and different stages werent showing stage name when in different axes [DHIS2-16462]

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-16462

In order to know if there is a duplicate it's necessary to check all the dimensions added to the layout, not just those in the same axis.

Also, instead of using `suffix`, or `stageName`, use the more generic `nameModifier`. This is in anticipation of the upcoming TEI multi-program implementation where the modifier could be a program name rather than a stage name.

Note that for the dimension modal, `stageName` is still used.

For the dragging chip, all the needed dimension info, including `nameModifier` is now provided in the dnd data object, so no need to find it in the metadata in DndContext.


jenniferarnesen:
This work is not currently prioritized. Keeping the branch for later.

PR #487, opened 2024-01-31
fix: when in the PeriodDimension, use the raw dimensionIds

jenniferarnesen:
Changes in this PR:
1. Fix the bug where the dimension items repeatedly get prefixed with the programId, such as: `programId.programId.programId.dimensionId`
2. Fix the bug where tooltip content was showing the metadata id rather than the name of the period. For the tooltip content, the metadata should just be THIS_MONTH, not programId.THIS_MONTH, since the periods will have the same name regardless of program.
3. The analytics query should add all the date conditions to a single string in the parameters list (see unit test). The only actual code change for this was the addition of this line: https://github.com/dhis2/line-listing-app/pull/487/files#diff-6ab44e5fc58d04348d840a9b86b9a5aa98750e705ae362de559fc5c926f21c5cR41
4. Add time dimension checks to existing cypress test.


cypress/helpers/period.js
```diff
@@ -7,11 +7,49 @@ const openPeriod = (label) => {
...
     openPeriod(label)
     cy.contains('Choose from presets').click()
     cy.contains('Fixed periods').click()
-    if (period.type) {
+
+    if (selected) {
+        // confirm that selected item is in the correct list in the transfer
+        if (
+            period.type &&
+            !cy
```
martinkrulltott, line 19: I'm not sure if you can actually do this (use the value returned from `cy` in an `if` statement) or if this just works by chance.
Regardless, I'd say given that `selected` is only used in one place, instead of making `selectFixedPeriod` overly complicated, just do everything manually directly in `createTrackedEntity.cy.js` instead. 
Preferably we keep the test files as simple as possible to avoid misuse and issues with the tests themselves (i.e. unlike "regular code", not everything has to go into a helper function for test code)

cypress/integration/createTrackedEntity.cy.js
```diff
@@ -75,37 +76,82 @@ const setUpTable = () => {
...
+    // Check the chip and tooltip content
+    cy.getBySelLike('layout-chip').contains(periodLabel).trigger('mouseover')
+    cy.getBySel('layout-chip-tooltip-content').contains(
+        `January ${getCurrentYearStr()}`
+    )
+    cy.getBySelLike('layout-chip').contains(periodLabel).trigger('mouseout')
+    cy.getBySel('layout-chip-tooltip-content').should('not.exist')
+
+    // Add another time dimension
+    // Check that the already selected period is not available to select
```
martinkrulltott, line 116: This should ideally checked after `Update` has been clicked and/or after the AO has been saved. That's when the metadata gets manipulated.

cypress/helpers/period.js
jenniferarnesen: Good catch. Code changed accordingly.

cypress/integration/createTrackedEntity.cy.js
```diff
@@ -75,37 +76,80 @@ const setUpTable = () => {
...
+    cy.getBySel('layout-chip-tooltip-content').should('not.exist')
 
     // Go back to person dimensions to verify that they're still listed properly
     cy.getBySel('main-sidebar').contains('Person dimensions').click()
     cy.getBySel('tracked-entity-dimensions-list')
         .children()
         .should('have.length', 33)
 
-    clickMenubarUpdateButton()
+    // clickMenubarUpdateButton()
```
martinkrulltott, line 148: This can be uncommented now that the backend is fixed? https://dhis2.atlassian.net/browse/DHIS2-16708

PR #534, opened 2024-06-03
feat: add resizable sidebar

HendrikThePendric:
Implements [DHIS2-XXXX](https://dhis2.atlassian.net/browse/DHIS2-XXXX)

**Requires https://github.com/dhis2/analytics/pull/XXX**

---

### Key features

1. _feature_

---

### Description

_text_

---

### TODO

-   [ ] Cypress tests
-   [ ] Update docs
-   [ ] Manual testing
-   [ ] _task_

---

### Known issues

-   [ ] _issue_

---

### Screenshots

_supporting text_


cypress/integration/mainSidebar.cy.js
```diff
@@ -1,9 +1,17 @@
+import {
+    ACCESSORY_PANEL_DEFAULT_WIDTH,
+    ACCESSORY_PANEL_MIN_WIDTH,
+    ACCESSORY_PANEL_MIN_PX_AT_END,
+    PRIMARY_PANEL_WIDTH,
+    ARROW_LEFT_KEY,
+    ARROW_RIGHT_KEY,
+} from '../../src/modules/accessoryPanelConstants.js'
```
HendrikThePendric, line 8: All these constants were moved to a separate file so they could also be imported into the e2e test files without causing transpilation errors

src/modules/accessoryPanelConstants.js
```diff
@@ -0,0 +1,12 @@
+export const ACCESSORY_PANEL_TAB_INPUT = 'INPUT'
+export const ACCESSORY_PANEL_TAB_PROGRAM = 'PROGRAM'
+export const ACCESSORY_PANEL_TAB_YOUR = 'YOUR'
+export const ACCESSORY_PANEL_TAB_TRACKED_ENTITY = 'TRACKED_ENTITY'
+export const ACCESSORY_PANEL_DEFAULT_WIDTH = 260
+export const ACCESSORY_PANEL_MIN_WIDTH = 180
+export const ACCESSORY_PANEL_MIN_PX_AT_END = 50
+export const PRIMARY_PANEL_WIDTH = 260
+export const ACCESSORY_PANEL_WIDTH_STORAGE_KEY =
```
edoardo, line 9: I wonder if we have some conventions about the naming.
In dashboard we have a couple of keys and the naming there is "prefixed" with`dhis2.dashboard.`.
I don't think it matters really, but we do seem to use different formats.

src/modules/ui.js
```diff
@@ -69,7 +75,23 @@ export const getDefaultSorting = () => ({
     direction: 'default',
 })
 
-export const ACCESSORY_PANEL_TAB_INPUT = 'INPUT'
-export const ACCESSORY_PANEL_TAB_PROGRAM = 'PROGRAM'
-export const ACCESSORY_PANEL_TAB_YOUR = 'YOUR'
-export const ACCESSORY_PANEL_TAB_TRACKED_ENTITY = 'TRACKED_ENTITY'
+export const getUserSidebarWidth = () => {
```
edoardo, line 78: In dashboard the code that deals with `localStorage` is in a separate file called `modules/localStorage.js`, but we have more keys there used in different places.

edoardo (approved):

src/modules/accessoryPanelConstants.js
HendrikThePendric: Perhaps it would be good to have a consistent approach across different apps, but I think that would be out-of-scope for this PR. Especially because quite a few of these constants were pre-existing, I added a few but others were only moved from one file to another. The ones that were added used the same naming conventions as the ones already present.

src/modules/ui.js
HendrikThePendric: Good point. I wasn't quite sure what to do with this when I wrote the code. I've implemented it in a separate file now, see [3023350](https://github.com/dhis2/line-listing-app/pull/534/commits/3023350ea8368d2b5c16b2938626701408cef633)

PR #572, opened 2024-08-07
fix: display the right format for time dimensions (DHIS2-17855)

edoardo:
Implements [DHIS2-17855](https://dhis2.atlassian.net/browse/DHIS2-17855)

---

### Key features

1. display the right format for time dimensions

---

### Description

From 42 all time dimensions are passed as `DATETIME` but with the exception of `lastUpdated` they don't have the time part set.
These time dimensions should be formatted as date.
For this we need a mapping in the LL app in order to override the `DATETIME` valueType and format the values correctly.

Some version checks in Cypress tests have been updated after the recent releases.

---

### TODO

-   [x] Cypress tests
-   [x] Manual testing

---

### Screenshots

Before:
Screenshot 2024-08-07 at 11 53 48

After:
Screenshot 2024-08-07 at 11 53 14

[DHIS2-17855]: https://dhis2.atlassian.net/browse/DHIS2-17855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/tableValues.js
+++ b/src/modules/tableValues.js
@@ -39,8 +39,9 @@ const getFormattedCellValue = ({ value, header = {}, visualization = {} }) => {
         return getStatusNames()[value] || value
     }
 
-    if ([VALUE_TYPE_DATE, VALUE_TYPE_DATETIME].includes(header.valueType)) {
-        let valueType = header.valueType
+    let valueType = header.valueType
+
+    if ([VALUE_TYPE_DATE, VALUE_TYPE_DATETIME].includes(valueType)) {
         if (
             header.name &&
             [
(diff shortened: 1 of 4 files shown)


src/modules/tableValues.js
```diff
@@ -29,11 +40,31 @@ const getFormattedCellValue = ({ value, header = {}, visualization = {} }) => {
     }
 
     if ([VALUE_TYPE_DATE, VALUE_TYPE_DATETIME].includes(header.valueType)) {
+        let valueType = header.valueType
```
janhenrikoverland, line 43: Maybe move this above the `if` and use it in the `includes` as well

janhenrikoverland: On line 72 I see `header?.valueType`, so maybe figure out if the `?` is needed everywhere or could be removed

janhenrikoverland, line 43: On line 72 I see `header?.valueType` - we should figure out if `?` is needed everywhere or can be removed

edoardo: I think it can be removed since `header` is initialised to empty object anyway on line 29.

edoardo: Removed.

janhenrikoverland (approved):

PR #576, opened 2024-08-12
fix: sanitize width before reading and writing to local storage

HendrikThePendric:
No JIRA ticket

---

### Key features

1. Ensures the width written to local storage is always an integer

---

### Description

I opted to sanitise both on read and write. This covers an edge case for users who have a broken app already, with an invalid width in local storage.

---

### TODO

-   [x] Manual testing

---

### DIFF
--- a/src/components/MainSidebar/MainSidebar.module.css
+++ b/src/components/MainSidebar/MainSidebar.module.css
@@ -65,7 +65,7 @@
 }
 .resizeHandle:hover::before {
     opacity: 1;
-    /* Show with a 500ms delay */
+    /* Show with a 200ms delay */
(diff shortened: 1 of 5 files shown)


janhenrikoverland (approved):

src/components/MainSidebar/MainSidebar.module.css
```diff
@@ -66,7 +66,7 @@
 .resizeHandle:hover::before {
     opacity: 1;
     /* Show with a 500ms delay */
```
HendrikThePendric, line 68:
```suggestion
    /* Show with a 200ms delay */
```

PR #581, opened 2024-08-23
fix: display correct app version in dhis2 and updated workflows

jenniferarnesen:
PR workflow: verify-pr.yml
* This workflow triggers on pull request actions only. It verifies that the PR code passes build, lint, test, and e2e.

Release workflow: release.yml
This workflow only runs on pushes to the master branch since this is the only release branch. Every release involves 2 commits to master:
* squash merging the PR
* The release workflow pushes a new commit with version change in package.json and CHANGELOG.md, e.g. the [skip release] commit

The release workflow triggers for both pushes to master listed above. But the release job should only run for the first push.

The semantic-release workflow steps are found in .releaserc. Previously @dhis2/action-semantic-release was used but this was overkill for our needs and difficult to make changes to. Therefore, we created our own semantic-release steps in .releaserc:
* commit-analyzer: Determines the next app version
* release-notes-generator: Generates the new release notes
* changelog: Updates CHANGELOG.md with the generated release notes
* npm: Updates package.json with the new version number. Does not publish to npm since private: true in package.json
* exec: Runs yarn build which updates all version number references in the bundle
* git: creates commit number "2" (mentioned above) with just CHANGELOG.md and package.json and pushes it to the remote
* github: pushes the bundle to the github repo as a new release and adds a tag

After semantic-release is complete, the bundle is then published to AppHub so that it is actually available to all users. Because of the addition of the exec yarn build step, all references to the version in the bundle should be the same, thus fixing the annoying bug where the old version displays in the instance after installing the later version.

If any point in the release fails, then a failure message is sent to slack. If the release step succeeds, then a success message is posted to slack that includes the new version number and the release notes.

Publish to d2-ci: publish-to-d2-ci.yml
This workflow publishes the build to d2-ci in the following cases:
* The [skip release] commit on master
* Git tags

Fail message is sent to slack if the publish failed


.github/workflows/publish-d2-ci.yml
```diff
@@ -0,0 +1,67 @@
...
+        if: ${{ failure() && !cancelled() }}
+        steps:
+            - name: Send failure message to analytics-internal-bot slack channel
+              id: slack_publish_failure
+              uses: slackapi/slack-github-action@v1.23.0
+              with:
+                  channel-id: ${{ secrets.SLACK_CHANNEL_ID }}
+                  payload: |
+                      {
+                          "text": ":small_red_triangle_down: :maps-app: Maps failed to deploy to d2-ci",
```
edoardo, line 64:
```suggestion
                          "text": ":small_red_triangle_down: :line-listing-app: Line Listing failed to deploy to d2-ci",
```

edoardo (approved):

PR #593, opened 2024-09-23
chore: use cypress-multi-reporters

adeldhis2:
Integrate cypress-multi-reporters to enable ReportPortal and spec reporting for improved GitHub Actions logging


package.json
```diff
@@ -30,12 +30,14 @@
...
         "@testing-library/react": "^12.1.4",
         "cypress": "^13.10.0",
+        "cypress-multi-reporters": "^1.6.4",
         "cypress-tags": "^1.2.2",
         "enzyme": "^3.11.0",
         "enzyme-adapter-react-16": "^1.15.8",
         "eslint-plugin-cypress": "^2.12.1",
         "identity-obj-proxy": "^3.0.0",
         "jest-enzyme": "^7.1.2",
+        "mochawesome": "^7.1.3",
```
HendrikThePendric, line 40: Just out of interest, why did we need this additional requirement? Is it required by report portal?

HendrikThePendric (approved):
This is great @adeldhis2, thanks a lot. I left one question about an additional dependency, but since it's a devDependency it's not a critical point.

package.json
adeldhis2: Good catch! I think this is not needed... let me remove it :)

jenniferarnesen:
Report portal has been removed so this PR no longer makes sense.

PR #651, opened 2025-02-27
Feat/allow multiple selection for org unit data elements dhis2 18560

HendrikThePendric:
Implements [DHIS2-XXXX](https://dhis2.atlassian.net/browse/DHIS2-XXXX)

**Requires https://github.com/dhis2/analytics/pull/XXX**

---

### Key features

1. _feature_

---

### Description

_text_

---

### TODO

-   [ ] Cypress tests
-   [ ] Update docs
-   [ ] Manual testing
-   [ ] _task_

---

### Known issues

-   [ ] _issue_

---

### Screenshots

_supporting text_


src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -73,6 +76,102 @@ const OrgUnitCondition = ({
...
+                !id.startsWith('USER_ORGUNIT')
+
+            if (isOuTreeItem) {
+                item.path = getOuPath(id, metadata, parentGraphMap)
+            }
+
+            return item
+        })
+    }, [condition, metadata, parentGraphMap])
+    const onSelect = useCallback(
```
jenniferarnesen, line 113: silly little comment, but could you add a newline, just for some breathing room in the file?

jenniferarnesen (approved):

src/components/App.js
```diff
@@ -213,7 +211,26 @@ const App = () => {
...
+                                data.eventVisualization.metaData
+                            )
+                        )
+                    )
+                    dispatch(
+                        acAddParentGraphMap(
+                            data.eventVisualization.parentGraphMap
+                        )
+                    )
+                }
```
HendrikThePendric, line 233: @jenniferarnesen I am updating the redux store conditionally here, because a change in the `metaData` state causes pretty much the full app to re-render. And `acAddMetadata` will produce a new `metaData` object even if we were to pass it `{}` or `undefined`. Since these properties don't exist pre-v42, I opted for an `if` here.

Even though this code doesn't technically branch on the version, I opted to add the `VERSION-TOGGLE` comment because this can be simplified in the future. I guess I am not strictly speaking the guidelines I outlined in the [FAQ](https://docs.google.com/document/d/1SnlYnXxWPps9y1jHwvAkaldokXVCWNIPFIFobLIoJs4/edit?tab=t.0#heading=h.508b6hssylfs), but this seemed like the most pragmatic solution for the current situation. Let me know if it makes sense to you.

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -72,6 +79,111 @@ const OrgUnitCondition = ({
         />
     )
 }
+const getMetaDataId = (id) =>
+    ouIdHelper.hasLevelPrefix(id) || ouIdHelper.hasGroupPrefix(id)
```
jenniferarnesen, line 83: You don't need to do these checks before calling `removePrefix` since that function does those checks internally.

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -72,6 +79,111 @@ const OrgUnitCondition = ({
...
+
+        return idsString.split(';').map((id) => {
+            const isOuTreeItem =
+                !ouIdHelper.hasLevelPrefix(id) &&
+                !ouIdHelper.hasGroupPrefix(id) &&
+                !id.startsWith('USER_ORGUNIT')
+            const metaDataId = getMetaDataId(id)
+            const item = {
+                id,
+                // We store IDs for levels and groups in the metadata store without prefix
```
jenniferarnesen, line 112: The comment doesn't seem necessary - code is pretty self explanatory

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -72,6 +79,111 @@ const OrgUnitCondition = ({
...
+            } else {
+                const { forMetadata, forParentGraphMap, itemIds } =
+                    items.reduce(
+                        (acc, item) => {
+                            const metaDataId = getMetaDataId(item.id)
+                            acc.itemIds.push(item.id)
+                            acc.forMetadata[metaDataId] = {
+                                id: metaDataId,
+                                name: item.name || item.displayName,
+                                displayName: item.displayName,
```
jenniferarnesen, line 138: is there any need for shortName support?

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -72,6 +79,111 @@ const OrgUnitCondition = ({
...
+                item.path = getOuPath(id, metadata, parentGraphMap)
+            }
+
+            return item
+        })
+    }, [condition, metadata, parentGraphMap])
+    const onSelect = useCallback(
+        ({ items }) => {
+            if (items.length === 0) {
+                onChange('')
```
jenniferarnesen, line 128: What scenarios would this happen in?

src/modules/conditions.js
```diff
@@ -21,8 +21,10 @@ import {
     VALUE_TYPE_ORGANISATION_UNIT,
     formatValue,
     DIMENSION_TYPE_PROGRAM_INDICATOR,
+    ouIdHelper,
 } from '@dhis2/analytics'
 import i18n from '@dhis2/d2-i18n'
+import metadata from '../reducers/metadata.js'
```
jenniferarnesen, line 27: lint error

src/modules/metadata.js
```diff
@@ -91,3 +91,37 @@ export const getDynamicTimeDimensionsMetadata = (
         return acc
     }, {}),
 })
+
+export const isPopulatedObject = (input) => {
```
jenniferarnesen, line 95: Could you add unit tests for both of these new functions?

src/components/Dialogs/Conditions/OrgUnitCondition.js
```diff
@@ -72,6 +79,111 @@ const OrgUnitCondition = ({
...
+    addMetadata,
+    addParentGraphMap,
+    condition,
+    onChange,
+    metadata,
+    parentGraphMap,
+}) => {
+    const { rootOrgUnits, currentUser } = useCachedDataQuery()
+    const selected = useMemo(() => {
+        const idsString = condition.split(':')?.[1] ?? ''
```
jenniferarnesen, line 98: It seems like we should have a function that does this, somewhere in LL, but since I'm not sure where, it is something we can improve later.

src/components/Dialogs/Conditions/OrgUnitCondition.js
HendrikThePendric: Yes perhaps the code is, but I still think this justifies some sort of comment, because of the following (copy-pasted from a Slack thread):
- When we query the backend, in the `dimension` query param, we need to use the IDs WITH the prefixes
- When we receive `metaData` from the backend we get plain IDs, WITHOUT the prefixes
- When we store things in the `metaData` redux store we do so using plain IDs, WITHOUT the prefixes
- When we interact with the `OrgUnitDimension` component (either providing the selected prop or when we read the arguments in the onSelected callback), we use the IDs WITH the prefixes

So it's not super straightforward 😅

src/components/Dialogs/Conditions/OrgUnitCondition.js
HendrikThePendric: Perhaps... But I'd classify it as out-of-scope for this PR because:
1. This isn't really new code, it's just been copied from the old component, and we weren't taking `shortName` into account before
2. If we tackle this here, we should really scan the full app for places where we write to the `metaData` store and ensure we take the `shortName` into account everywhere

Shall we create a JIRA ticket for this?

src/components/Dialogs/Conditions/OrgUnitCondition.js
HendrikThePendric: This is not hypothetical, it happens when the only one item is selected and it then gets deselected. An when this happens we do need to call `onChange` with an empty string, and not with the `IN` operator.

src/components/Dialogs/Conditions/OrgUnitCondition.js
HendrikThePendric: Good find, but it turns out I can't use this `parseCondition` helper. The reason is related to my previous [comment here](https://github.com/dhis2/line-listing-app/pull/651#discussion_r2000593435). If `condition` is an empty string, then something unexpected happens: `parseCondtion('')` returns `[ '' ]` and we then the map callback fails.

I suppose I could look into fixing `parseCondition`, because `''` should return `[]` and not `[ '' ]`. But I guess this could have some unintended consequences.

What do you think?

src/components/Dialogs/Conditions/OrgUnitCondition.js
jenniferarnesen: As in, it should work to just have:

`getMetaDataId = id => ouIdHelper.removePrefix(id)`

src/components/Dialogs/Conditions/OrgUnitCondition.js
jenniferarnesen: yeah, let's just wait on it. Is there a way to flag this with a // REFACTOR or something without triggering a new code smell?

src/modules/metadata.js
HendrikThePendric: Yes will do. Actually `transformMetaDataResponseObject` is not really new. This is simply some logic that was extracted from the component because we now need to do the same thing in two places.

src/components/Dialogs/Conditions/OrgUnitCondition.js
jenniferarnesen: I'll leave it up to you. Just pointing out that the current comment basically just reiterates exactly what the code says. Maybe this deserves more of a global comment? I know we are stripping prefixes for the metadata elsewhere in the code. (FixedDimension.js)

src/components/Dialogs/Conditions/OrgUnitCondition.js
jenniferarnesen: I'll check into what Jira tickets we have already for shortName.

src/components/Dialogs/Conditions/OrgUnitCondition.js
HendrikThePendric: I'll just remove the comment

PR #673, opened 2025-04-11
fix: handle empty createdby and lastupdatedby

janhenrikoverland:
TODO

### DIFF
--- a/src/modules/tableValues.js
+++ b/src/modules/tableValues.js
@@ -15,6 +15,8 @@ import {
     DIMENSION_ID_INCIDENT_DATE,
     DIMENSION_ID_SCHEDULED_DATE,
     DIMENSION_ID_LAST_UPDATED,
+    DIMENSION_ID_CREATED_BY,
+    DIMENSION_ID_LAST_UPDATED_BY,
 } from './dimensionConstants.js'
 import { extractDimensionIdParts } from './dimensionId.js'
 import { getTimeDimensions } from './timeDimensions.js'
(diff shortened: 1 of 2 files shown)


src/modules/tableValues.js
```diff
@@ -39,6 +41,17 @@ const getFormattedCellValue = ({ value, header = {}, visualization = {} }) => {
         return getStatusNames()[value] || value
     }
 
+    // TODO: Remove this when the problem with empty/partial user info is resolved in backend
```
jenniferarnesen, line 44: This comment implies there is an open Jira issue for the empty/partial user info. Could you include the Jira issue number in the comment?

src/modules/tableValues.js
```diff
@@ -39,6 +41,17 @@ const getFormattedCellValue = ({ value, header = {}, visualization = {} }) => {
...
+    if (
+        [
+            headersMap[DIMENSION_ID_CREATED_BY],
+            headersMap[DIMENSION_ID_LAST_UPDATED_BY],
+        ].includes(header.name) &&
+        value === ',  ()'
```
jenniferarnesen, line 50: Here are some tests you can add to tableValues.spec.js:

```
// TODO DHIS2-XXXXX
describe('partial user info bug workaround', () => {
        it("returns empty string when createdBy or lastUpdatedBy is ', ()'", () => {
            const header = { name: 'createdbydisplayname' }
            const value = ',  ()' // 2 spaces between the comma and parentheses

            expect(getFormattedCellValue({ header, value })).toBe('')
        })
        it("returns empty string when createdBy or lastUpdatedBy is ', ()'", () => {
            const header = { name: 'lastupdatedbydisplayname' }
            const value = ',  ()' // 2 spaces between the comma and parentheses

            expect(getFormattedCellValue({ header, value })).toBe('')
        })
    })
```

PR #680, opened 2025-04-24
chore: fix test visualizations cleanup in Cypress

edoardo:
### Description

The previous PR which implemented the cleanup of all created visualization during Cypress test runs had a bug in the extraction of the ids which prevented those visualizations from being deleted.
The regexp is fixed here to correctly extract the id from the URL.

All the names used for the saved visualization have been shortened and they should be unique.
The full date and time is being replaced with the epoch.

---

### Quality checklist

Add _N/A_ to items that are not applicable.

- [ ] Dashboard tested N/A
- [x] Cypress and/or Jest tests added/updated
- [ ] Docs added N/A
- [ ] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX) N/A
- [ ] Tester approved (name) N/A

### DIFF
--- a/cypress/integration/fileMenu.cy.js
+++ b/cypress/integration/fileMenu.cy.js
@@ -166,7 +166,7 @@ describe('file menu', () => {
 
         clickMenubarUpdateButton()
 
-        const AO_NAME = `TEST ${new Date().toLocaleString()} - "saved, valid: save" state`
+        const AO_NAME = `TEST-${Date.now()}-valid-save`
 
         saveVisualization(AO_NAME)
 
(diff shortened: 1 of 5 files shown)


cypress/integration/fileMenu.cy.js
```diff
@@ -166,7 +166,7 @@ describe('file menu', () => {
 
         clickMenubarUpdateButton()
 
-        const AO_NAME = `TEST ${new Date().toLocaleString()} - "saved, valid: save" state`
+        const AO_NAME = `TEST-${Date.now()}-valid-save`
```
jenniferarnesen, line 169: I like how you've set the name to reflect the test file. Maybe do the same here and in Save?

jenniferarnesen (approved):

PR #681, opened 2025-04-25
fix: fetch original visualization before renaming so as not to overwrite newer changes

jenniferarnesen:
Fixes [DHIS2-19506](https://dhis2.atlassian.net/browse/DHIS2-19506)

Description
Rename: fetch the visualization including subscribers, and apply name and description to that object, which is then used in the rename PUT request. In the app, only name and description are updated, which means the app's copy of the visualization could still be out of date. But that's a long-standing problem that will have to be addressed later.

Save: fetch just the visualization subscribers and apply that, to the changed visualization.

Load: when loading, do not include subscribers in the visualization request because it is not in use. This also prevents SaveAs from having to delete the subsribers object.

Quality checklist
- [ ] Dashboard tested N/A
- [x] Tests added (Cypress and/or Jest)
- [ ] Docs added N/A
- [ ] Global shell tested N/A
- [x] d2-ci dependency replaced https://github.com/dhis2/analytics/pull/1777
- [x] Tester sign-off (Edoardo)

To DO
- [x] Restore original package.json version

[DHIS2-19506]: https://dhis2.atlassian.net/browse/DHIS2-19506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "line-listing-app",
-    "version": "999.99.0-rename.june2",
+    "version": "102.1.8",
     "description": "DHIS2 Line Listing",
     "license": "BSD-3-Clause",
     "private": true,


edoardo (approved):

src/api/visualization.js
```diff
@@ -0,0 +1,101 @@
...
+                '*',
+                `columns[${dimensionFields}]`,
+                `rows[${dimensionFields}]`,
+                `filters[${dimensionFields}]`,
+                `program[id,programType,${nameProp}~rename(name),displayEnrollmentDateLabel,displayIncidentDateLabel,displayIncidentDate,programStages[id,displayName~rename(name),repeatable]]`,
+                'programStage[id,displayName~rename(name),displayExecutionDateLabel,displayDueDateLabel,hideDueDate,repeatable]',
+                `programDimensions[id,${nameProp}~rename(name),enrollmentDateLabel,incidentDateLabel,programType,displayIncidentDate,displayEnrollmentDateLabel,displayIncidentDateLabel,programStages[id,${nameProp}~rename(name),repeatable,hideDueDate,displayExecutionDateLabel,displayDueDateLabel]]`,
+                'access',
+                'href',
+                !withSubscribers && '!subscribers',
```
edoardo, line 22: This ends up adding 2 `false` entries in the list of fields:
Screenshot 2025-05-22 at 16 38 00

edoardo (commented):
Same considerations as in the DV PR about the regexp in the Cypress tests.

There is a problem with the conditional addition of `subscribers` and `subscribed`, see the comment.

src/api/visualization.js
```diff
@@ -0,0 +1,103 @@
...
+                '!dataElementOperands',
+                '!dataElementGroups',
+                '!dataSets',
+                '!periods',
+                '!organisationUnitLevels',
+                '!organisationUnits',
+                '!user',
+            ]
+            if (!withSubscribers) {
+                fields.push('!subscribers')
```
edoardo, line 52: Shouldn't this have `subscribed` too?
```suggestion
                fields.push('!subscribers','!subscribed')
```
If it's not used at all it can be added outside the if block.

cypress/integration/save.cy.js
```diff
@@ -210,16 +82,73 @@ describe('save', () => {
...
+        // check user is still subscribed
+        cy.getBySel('dhis2-analytics-interpretationsanddetailstoggler').click()
+        cy.get('button').contains('Unsubscribe').should('be.visible')
+        cy.getBySel('dhis2-analytics-interpretationsanddetailstoggler').click()
+        cy.contains('About this line list').should('not.exist')
+
         // save as with name change
+        cy.intercept(
+            'POST',
+            /\/api\/\d+\/eventVisualizations(\?.*)?/,
```
edoardo, line 137: And here

edoardo (dismissed):

edoardo (approved):

